PR #3, opened 2019-04-18
RIUI-1724 Logger implemented 

dpanov76:


src/environments/environment.ts
```diff
@@ -1,9 +1,11 @@
 // This file can be replaced during build by using the `fileReplacements` array.
 // `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
 // The list of file replacements can be found in `angular.json`.
+// isDebugMode from 0 to 3 when 0 then logger completely disabled
 
 export const environment = {
-  production: false
+  production: false,
+  isDebugMode: 3
```
matthewspeake-hmcts: loggingLevel would be a better name for this.  isXXX implies a boolean.

matthewspeake-hmcts: Also, is there a way of maintaining a map for a meaningful level to map what you showed yesterday:

1=ERROR
2=WARN
3=INFO
4=DEBUG

so we could set it like this:

```
loggingLevel: 'WARN'
```

or 

```
loggingLevel: Level.WARN
```

matthewspeake-hmcts:
We can do those changes I mentioned in another ticket, for now it's good to go! 👍 Nice work! ;)

src/environments/environment.ts
```diff
@@ -1,9 +1,11 @@
 // This file can be replaced during build by using the `fileReplacements` array.
 // `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
 // The list of file replacements can be found in `angular.json`.
+// isDebugMode - 'INFO' 'WARN 'ERROR'
 
 export const environment = {
```
frontr-uk: create interface: 

export interfaces  ENVIROMENT {
 production: boolean,
isDebugMode: 'INFO', 'WARN', 'ERROR'
}

export const enviroment: ENVIROMENT {

}

frontr-uk: this way we make sure the user does not accidentally put a wrong value in the property.

PR #17, opened 2019-05-10
Search filter RIUI-1760

dpanov76:


src/cases/containers/cases-create/cases-create.component.ts
```diff
@@ -50,4 +107,22 @@ export class CasesCreateComponent {
...
   }
 
+  // for search results
+
+  ngOnInit() {
+    this.jurisdictions = [this.JURISDICTION_1, this.JURISDICTION_2];
+  }
+
+  applied(selected) {
+    console.log('selected:', selected);
```
fazanki: applied, reset and jusristictionSelected will get picked up by connector component

Instead what we should be providing is biding array that binds those events to exui actions to connector component

same as on case create

src/cases/containers/cases-create/cases-create.component.ts
```diff
@@ -32,6 +41,54 @@ export class CasesCreateComponent {
 
   caseSelected: string;
 
+
+  //For Searchfilters only
```
fazanki: i think we need to see about this

dpanov76: Rebased please review again

PR #19, opened 2019-05-13
Feature/production ready node

smogese:
This includes code that  supports  authentication within local development and code that serves code from the dist folder under one combined server to be used within the build pipeline

This also includes rudimentary proxy support which allows component rendering


package.json
```diff
@@ -5,15 +5,19 @@
     "nuke": "rm -rf dist",
     "build": "npm run build:ng && npm run build:node",
     "pre-push": "npm run lint && ng test --code-coverage && npm run build:prod",
-    "start": "node ./dist/rpa-pui-registration/api/server.js",
+    "start": "node ./dist/rpx-exui/api/server.js",
```
fazanki: can you update the pre-push lint so to lint:ng

PR #24, opened 2019-05-16
added mocha framework

smogese:
Unit tests for node

### DIFF
--- a/api/lib/middleware/auth.spec.ts
+++ b/api/lib/middleware/auth.spec.ts
@@ -14,7 +14,7 @@ import * as idam from '../../services/idam'
 import * as auth from './auth'
 
 describe('auth', () => {
-    // commenting out as not clear if restricted roles apply to xui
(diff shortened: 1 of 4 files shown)


api/lib/middleware/auth.spec.ts
```diff
@@ -14,62 +14,63 @@ import * as idam from '../../services/idam'
...
-            expect(auth.validRoles(['jui-judge'])).to.equal(true)
-            expect(auth.validRoles(['jui-panelmember'])).to.equal(true)
-        })
-
-        xit('should return false if given roles do not match jui pannel member or judge roles', () => {
-            expect(auth.validRoles(['test'])).to.equal(false)
-        }) // FIXME
-
-    })
+    // commenting out as not clear if restricted roles apply to xui
```
fazanki: is this something that will be addressed later on ? maybe TODO

smogese: Good point , i'll make it a TODO

PR #27, opened 2019-05-17
EUI-56 search filter is unit tests

dpanov76:


src/cases/store/reducers/search-filter.reducer.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  describe('APPLIED action', () => {
+    it('should apply patload to store', () => {
+      const { initialStateSearch } = fromCases;
+      const payload = {};
+      const action = new fromActions.Applied(payload);
+      const state = fromCases.reducer(initialStateSearch, action);
+      expect(state.loading).toEqual(false);
+      expect(state.loaded).toEqual(false);
+    });
```
fazanki: can you add test for serachFilterCiteria property and the loaded state should change to true

PR #30, opened 2019-05-20
enable signout via ngrx action

masahmad:

https://tools.hmcts.net/jira/browse/EUI-42

### Change description ###

code change done

```
[ ] Yes
[x] No
```


src/app/services/auth/auth.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  constructor(
+    public authService: AuthService,
+  ) {
+  }
+
+
+  canActivate() {
+    console.log('reached can activate');
+    if (!this.authService.isAuthenticated()) {
```
fazanki: maybe you can remove this

src/app/store/reducers/app-config.reducer.ts
```diff
@@ -24,10 +24,16 @@ export function reducer(
         loaded: true
       };
     }
+    case fromActions.LOGOUT: {
+      return {
+        ...state,
```
fazanki: maybe do to do or something because atm is not doing anything

src/app/services/auth/auth.guard.ts
masahmad: removed

src/app/store/reducers/app-config.reducer.ts
masahmad: agree

PR #33, opened 2019-05-23
Nojia/terraform set asp name

matthewspeake-hmcts:
PR to set asp name


matthewspeake-hmcts:
Approved by @smogese after slack chat.

PR #35, opened 2019-05-23
Apply contained CCD component's tagname to container's data-selector

serlingpa:
https://tools.hmcts.net/jira/browse/EUI-131

This branch adds a `data-selector` attribute to the connector component's element with a value of the tag name of the contained CCD component.

### DIFF
--- a/src/app/containers/exiu-ccd-connector-wrapper/exui-ccd-connector.component.ts
+++ b/src/app/containers/exiu-ccd-connector-wrapper/exui-ccd-connector.component.ts
@@ -1,6 +1,14 @@
-import {AfterContentInit, Component, ContentChild, Input, OnDestroy} from '@angular/core';
-import {Subscription} from 'rxjs';
-import {Store} from '@ngrx/store';
+import {
+  AfterContentInit,
+  Component,
+  ContentChild,
+  ElementRef, HostBinding,
+  Input,
+  OnDestroy
+} from '@angular/core';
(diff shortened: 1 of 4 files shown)


Mo-Lala84:
Approved. Good work Mark

src/app/containers/exiu-ccd-connector-wrapper/exui-ccd-connector.component.ts
```diff
@@ -17,14 +25,15 @@ import {Store} from '@ngrx/store';
   `
 })
 export class ExuiCcdConnectorComponent implements AfterContentInit, OnDestroy {
-
   @ContentChild('ccdComponent') ccdComponent;
+  @ContentChild('ccdComponent', {read: ElementRef}) ccdComponentElementRef: ElementRef;
```
fazanki: you got double ContentChild decorator... no need to reference it twice. Think you should be able to extend line 28

fazanki:
would be good to update comments to match your extension and add a little unit test for hostBindingValue

src/app/containers/exiu-ccd-connector-wrapper/exui-ccd-connector.component.ts
serlingpa: Unfortunately, we need both. We need both a reference to both the HTML element and the component instance itself.

Mo-Lala84:
Fine to commit :-)

PR #66, opened 2019-06-10
Bug/add missing preview config for prs

smogese:
Now that pipeline ENVS are being set up correctly, PRS don't just revert to 'local' which uses aat, but read the env setting, which is 'preview'. However no code is in place for this and never was. Things were working due to coincidence.


api/config/index.ts
```diff
@@ -4,10 +4,12 @@ import { application } from './application.config'
 
 import * as aat from './environments/aat.config'
 import * as local from './environments/local.config'
+import * as preview from './environments/preview.config'
 
 export const configs = {
     aat,
     local,
+    preview,
```
toireasam: will comma on end of this throw exception? should prob be removed

smogese: nope that's the standard for node es6/typecript linting
https://eslint.org/docs/rules/comma-dangle

PR #73, opened 2019-06-13
Add footer to application

borislavpetrovhmcts:
Adds footer to application

Fixes issue EUI-214


src/app/containers/footer/footer.component.ts
```diff
@@ -0,0 +1,30 @@
...
+        },
+        phone: {
+            text: '0207 633 4140'
+        },
+        opening: {
+            text: 'Monday to Friday, 8am to 6pm (excluding public holidays)'
+        }
+    };
+    public navigationData: Navigation = {
+        items: [
```
smogese: Should we have something in place for configuration data so it's not inline with code?

borislavpetrovhmcts: This is now resolved. Footer content exported.

PR #76, opened 2019-06-14
Chore/fix preview env

smogese:

### Change description ###

```
[ ] Yes
[ ] No
```


api/services/idam.ts
```diff
@@ -41,6 +43,7 @@ export async function getUser(email = null) {
 
 export async function postOauthToken(code, host) {
     const redirectUri = `${idamProtocol}://${host}/${oauthCallbackUrl}`
+    logger.info(`idam details ${idamClient}:${idamSecret}`)
```
andywilkinshmcts: Is it a good idea to log idam secrets? Especially at info level.

andywilkinshmcts: I see you removed it in later commit

smogese: Its fine until it goes into prod from a sec point of view, but these details were only put into help debug what was going on (very usefully too), so I removed them in subsequent commits.

PR #140, opened 2019-09-11
scope values added EUI-767

Vamshi-Muniganti:

https://tools.hmcts.net/jira/browse/EUI-767

### Change description ###

adding scope id values

```
[ ] Yes
[ x] No
```


Vamshi-Muniganti:
> we don't need scope 'manage'roles' as we are not editing roles for manage cases.

manage roles is removed from scope id

PR #162, opened 2019-09-17
title update on manage case

praveenrajendran101:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/cases/containers/case-create/case-create.component.ts
+++ b/src/cases/containers/case-create/case-create.component.ts
@@ -5,6 +5,7 @@ import * as fromCases from '../../../cases/store';
 import * as fromRoot from '../../../app/store';
 import {ActionBindingModel} from '../../models/create-case-actions.model';
 import {Subscription} from 'rxjs';
+import { Title } from '@angular/platform-browser';
 /**
  * Entry component wrapper for CddCreateCaseFilter
  * Smart Component consuming <ccd-case-create>
@@ -22,7 +23,9 @@ export class CasesCreateComponent implements OnInit, OnDestroy {
   fromCasesFeature: any;
   $inputSubscription: Subscription;
 
-  constructor(private store: Store<fromCaseCreate.State>) {}
+  constructor(private store: Store<fromCaseCreate.State>, private titleService: Title) {
+    this.titleService.setTitle('HMCTS Manage Cases | Create a case');
+  }
 
   ngOnInit(): void {
     this.fromCasesFeature = fromCases;
--- a/src/cases/containers/case-create/case-create.component.ts
+++ b/src/cases/containers/case-create/case-create.component.ts
@@ -5,7 +5,6 @@ import * as fromCases from '../../../cases/store';
 import * as fromRoot from '../../../app/store';
 import {ActionBindingModel} from '../../models/create-case-actions.model';
 import {Subscription} from 'rxjs';
-import { Title } from '@angular/platform-browser';
 /**
  * Entry component wrapper for CddCreateCaseFilter
  * Smart Component consuming <ccd-case-create>
--- a/src/cases/containers/case-create/case-create.component.ts
+++ b/src/cases/containers/case-create/case-create.component.ts
@@ -23,8 +23,7 @@ export class CasesCreateComponent implements OnInit, OnDestroy {
   fromCasesFeature: any;
   $inputSubscription: Subscription;
 
-  constructor(private store: Store<fromCaseCreate.State>, private titleService: Title) {
-    this.titleService.setTitle('HMCTS Manage Cases | Create a case');
+  constructor(private store: Store<fromCaseCreate.State>) {
   }
 
   ngOnInit(): void {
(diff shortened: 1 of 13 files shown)


praveenrajendran101:
> Not overly keen on this approach, yes it works but means having to inject the TitleService everywhere. Perhaps a better approach would be to use route data and maybe another interceptor/service/component somewhere? This also means the titles are then solely tied to the routes themselves and not individual components. See for an (outdated) example https://ultimatecourses.com/blog/dynamic-page-titles-angular-2-router-events

> This is solely based on the common text in the titles, if the BA's comeback with a different title later on this implementation wont work

paddybasi:
If that ever becomes the case:
1) Implementing it as an interceptor means the common part of the title (e.g. 'HMCTS Manage Cases |') can be abstracted away. Would therefore require less change.
2) In the 'edge case' where we can't dynamically set the title from the route, at this point we can inject the titleService and manually set it

paddybasi:
Also, have a look at app.constants.ts if we don't want to tie the titles to the routes directly

praveenrajendran101:
This is not edge case, case module is a dynamically loaded module. You are mentioning 'HMCTS Manage Cases |' in the root of the solution and the rest in routes of dynamically loaded cases module. Architecturally, it is like text splitting without any base component from root solution implemented in Cases module. Hence, splitting data between modules is no a good idea.

src/cases/containers/case-search/case-search.component.ts
```diff
@@ -6,6 +6,7 @@ import { Observable, combineLatest, Subscription } from 'rxjs';
 import { AppConfig } from '../../../app/services/ccd-config/ccd-case.config';
 import { ActionBindingModel } from '../../../cases/models/create-case-actions.model';
 import { FormGroup } from '@angular/forms';
+import { Title } from '@angular/platform-browser';
```
paddybasi: clean up

src/cases/containers/case-list/case-list.component.ts
```diff
@@ -60,7 +61,10 @@ export class CaseListComponent implements OnInit, OnDestroy {
     public store: Store<fromCaseList.State>,
     private appConfig: AppConfig,
     private definitionsService: DefinitionsService,
-  ) { }
+    private titleService: Title
```
paddybasi: This still needed here?

src/cases/containers/case-home/case-home.component.ts
```diff
@@ -1,7 +1,10 @@
+import { ExUITitleService } from 'src/app/shared/services/exui-title.service';
 import { Component } from '@angular/core';
 
 @Component({
   selector: 'exui-case-home',
   templateUrl: 'case-home.component.html'
 })
-export class CaseHomeComponent {}
+export class CaseHomeComponent {
```
paddybasi: clean up

src/cases/containers/case-filter/case-filter.component.ts
```diff
@@ -4,6 +4,7 @@ import {Store} from '@ngrx/store';
 import * as fromCaseCreate from '../../store/reducers';
 import {ActionBindingModel} from '../../models/create-case-actions.model';
 import * as fromCasesFeature from '../../store';
+import { Title } from '@angular/platform-browser';
```
paddybasi: clean up

src/cases/containers/case-details-home/case-details-home.ts
```diff
@@ -0,0 +1,10 @@
+import { Title } from '@angular/platform-browser';
```
paddybasi: clean up

src/cases/containers/case-create/case-create.component.ts
```diff
@@ -5,6 +5,7 @@ import * as fromCases from '../../../cases/store';
 import * as fromRoot from '../../../app/store';
 import {ActionBindingModel} from '../../models/create-case-actions.model';
 import {Subscription} from 'rxjs';
+import { Title } from '@angular/platform-browser';
```
paddybasi: clean up

src/cases/containers/case-search/case-search.component.ts
praveenrajendran101: done

src/cases/containers/case-list/case-list.component.ts
praveenrajendran101: done

src/cases/containers/case-home/case-home.component.ts
praveenrajendran101: done

src/cases/containers/case-filter/case-filter.component.ts
praveenrajendran101: done

src/cases/containers/case-details-home/case-details-home.ts
praveenrajendran101: done

src/cases/containers/case-create/case-create.component.ts
praveenrajendran101: done

PR #170, opened 2019-09-23
Open files inline

paddybasi:
https://tools.hmcts.net/jira/browse/EUI-818

### Change description ###
Added missing headers. Headers are sent in node response layer as values from response in proxy to dm-store. Added tests

```
[ ] Yes
[x] No
```


ficklephil:
Looks good!

PR #222, opened 2019-10-14
App utils tests.

ficklephil:
App Utils Tests

### DIFF
--- a/src/app/app-utils.spec.ts
+++ b/src/app/app-utils.spec.ts
@@ -1,0 +1,40 @@
+import {AppConstants} from './app.constants';
+
+import {AppUtils} from './app-utils';
+
+describe('getEnvironment', () => {
+
+  it('should return the prod environment for a blank url.', () => {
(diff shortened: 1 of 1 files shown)


src/app/app-utils.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it('should return the ithc environment for a ithc url.', () => {
+    expect(AppUtils.getEnvironment('ithc')).toEqual(
+      AppConstants.ENVIRONMENT_NAMES.ithc);
+  });
+
+  it('should return the perftest environment for a perftest url.', () => {
+    expect(AppUtils.getEnvironment('perftest')).toEqual(
+      AppConstants.ENVIRONMENT_NAMES.perftest);
+  });
```
praveenrajendran101: It needs a test for wrong url. It seems if you input wrong url to AppUtils.getEnvironment and if it is not matched any of the urls, it will return production url. It doesn't serve the purpose.Hence there should be a test with wrong url.

PR #265, opened 2019-11-04
adding filetes for returned jurisdiction

fazanki:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/api/amendedJurisdictions/index.ts
+++ b/api/amendedJurisdictions/index.ts
@@ -4,25 +4,26 @@ import { config } from '../config'
 import { http } from '../lib/http'
 import { setHeaders } from '../lib/proxy'
 
+/**
+ * Manually filtering returned jurisdictions
+ * to make available jurisdiction in filters array only
+ */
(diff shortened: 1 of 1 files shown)


api/amendedJurisdictions/index.ts
```diff
@@ -4,25 +4,26 @@ import { config } from '../config'
...
+    console.log(req.url)
     let url = striptags(req.url)
+    console.log(url)
     url = req.baseUrl  + url
     const headers: any = setHeaders(req)
 
     try {
         const response = await http.get(`${config.services.ccd.componentApi}${url}`, { headers })
-
+        const filters = {jurisdiction: ['DIVORCE', 'PROBATE']}
```
adnanakgun: Is the extra `jurisdiction` layer necessary? `filters` could be an array instead I think.

fazanki: hmm dunno what thinking that it can be filtered by another thing in the future... what do u think?

adnanakgun: I think there won't be a need for that, but I see your point.

PR #293, opened 2019-11-21
Bug/eui 1115 case progression aa

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-1115

### Change description ###

Copied and adapted navigation handler from CCD Consumer

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/app/store/effects/router.effect.ts
+++ b/src/app/store/effects/router.effect.ts
@@ -23,7 +23,7 @@ export class RouterEffects {
     ofType(RouterActions.GO),
     map((action: RouterActions.Go) => action.payload),
     tap(({ path, query: queryParams, extras, callback, errorHandler }) => {
-      this.router.navigate(path, { queryParams, ...extras })
+      return this.router.navigate(path, { queryParams, ...extras })
         .then(() => callback ? callback() : false)
         .catch(error => errorHandler ? errorHandler(error) : false);
     })
@@ -35,7 +35,7 @@ export class RouterEffects {
     map((action: RouterActions.CreateCaseGo) => action.payload),
     tap(({ path, query: queryParams, extras, caseId }) => {
       const thatCaseId = caseId;
-      this.router.navigate(path, { queryParams, ...extras }).then(() => {
+      return this.router.navigate(path, { queryParams, ...extras }).then(() => {
         this.store.dispatch(new fromCases.CreateCaseLoaded(thatCaseId));
       });
     })
--- a/src/app/store/effects/router.effect.ts
+++ b/src/app/store/effects/router.effect.ts
@@ -22,8 +22,10 @@ export class RouterEffects {
   navigate$ = this.actions$.pipe(
     ofType(RouterActions.GO),
     map((action: RouterActions.Go) => action.payload),
-    tap(({ path, query: queryParams, extras }) => {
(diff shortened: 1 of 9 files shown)


src/app/store/effects/router.effect.ts
```diff
@@ -22,8 +22,10 @@ export class RouterEffects {
   navigate$ = this.actions$.pipe(
     ofType(RouterActions.GO),
     map((action: RouterActions.Go) => action.payload),
-    tap(({ path, query: queryParams, extras }) => {
-      this.router.navigate(path, { queryParams, ...extras });
+    tap(({ path, query: queryParams, extras, callback, errorHandler }) => {
```
fazanki: have learned lately that tap operator is usually the best used only for debugging purposes i.e to check memory leeks. you can use another map - also this observable is not completed as is not returning router.navigate

src/cases/containers/case-home/case-home.component.ts
```diff
@@ -1,10 +1,113 @@
...
+    private titleService: ExUITitleService,
+    private alertService: AlertService,
+    private navigationNotifier: NavigationNotifierService,
+    private store: Store<fromFeature.State>,
+  ) { }
+
+  ngOnInit(): void {
+    this.navigationSubscription = this.navigationNotifier.navigation.subscribe(navigation => {
+      if (navigation.action) {
+        this.actionDispatcher(this.paramHandler(navigation));
```
fazanki: this should move trigger effects

src/cases/containers/case-home/case-home.component.ts
```diff
@@ -1,10 +1,113 @@
...
+
+  }
+
+  ngOnDestroy(): void {
+    if (this.navigationSubscription) {
+      this.navigationSubscription.unsubscribe();
+    }
+  }
+
+  paramHandler(navigation: any): GoActionParams {
```
fazanki: this should move into effects and dispatch the .Go action directly within the statements

src/app/store/effects/router.effect.ts
adnanakgun: Can you redirect me to a source regarding tap-debugging please? What should we use instead?

The `return` makes sense, I'll fix that.

src/cases/containers/case-home/case-home.component.ts
adnanakgun: I'm not quite clear what you mean with this. The component only needs to trigger the `Go` action. The parameters and callbacks/error-handlers are prepared and passed on to the action. I don't understand where or why there is a need for an effect.

src/cases/containers/case-home/case-home.component.ts
adnanakgun: Same as above.

src/app/store/effects/router.effect.ts
fazanki: the source i have is in a book not with me.... i can explain tomorrow why is not a good idea... something to do with side effects... as if you resue the stream somewhere else i bring that side effect with it...

src/cases/containers/case-home/case-home.component.ts
fazanki: yeah fine but is a side effect and needs to be managed in the effect... so is just about to move this code to the effects

PR #307, opened 2019-12-04
AWAITING CAB: Stop a User accessing ManageCases if they do not have a caseworker role.

ficklephil:
Restrict access to Manage Cases if they do not have a caseworker role as per EUI-1073.

### DIFF
--- a/api/auth/manageCasesUserRoleAuth.ts
+++ b/api/auth/manageCasesUserRoleAuth.ts
@@ -1,4 +1,15 @@
-export const CASEWORKER_INDENTIFIER = 'caseworker'
+import {config} from '../config'
+
+/**
+ * getLoginRoleMatcher
+ *
+ * The role who is able to access this application should have a prefix of 'caseworker'.
+ *
+ * The matcher is passed in through the config.
+ *
+ * @returns {string} - ie. 'caseworker'
+ */
+export const getLoginRoleMatcher = () => config.loginRoleMatcher
 
 /**
  * User Has Application Access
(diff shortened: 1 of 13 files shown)


api/auth/manageCasesUserRoleAuth.ts
```diff
@@ -0,0 +1,15 @@
+export const CASEWORKER_INDENTIFIER = 'caseworker'
```
paddybasi: In 12FA, maybe we should extract this into an environment variable?

Mo-Lala84: We need abstract static data out in env var into our config

borislavpetrovhmcts: The 'casworker' role does not change with environments, why extracting it to an   environment variable ?

ficklephil: I don't mind changing it, but we currently have probably around +400 users in the wild with 'caseworker-' roles. I would argue that it might be better to keep it here; as it's unit testable therefore we can make sure it doesn't change from 'caseworker'. If we have it as a configuration parameter, and someone accidentally changes it, it would mean that no one login, therefore I believe that having it as a config parameter increases our risk. 

But then if the configuration parameters are locked down on Production it should be fine.

ficklephil: 'caseworker' is now coming from the configuration parameters.

api/lib/objectUtilities.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+describe('Object Utilities ', () => {
+
+  describe('propsExist()', () => {
+
+    it('Should return true if all the properties exist on an object.', () => {
+
+      const object = {level1: {level2: {level3: 'level3'}}}
+
+      expect(propsExist(object, ['level1', 'level2', 'level3'])).to.be.true
```
paddybasi: So, we could do expect(exists(object, 'level1.level2.level3')).to.be.true

api/lib/objectUtilities.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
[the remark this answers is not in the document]
ficklephil: Discussed with Paddy

PR #337, opened 2020-01-14
Bug/eui 1342 create case cancel link

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-1342

### Change description ###

Added cancel effect 

```
[ ] Yes
[x] No
```


src/cases/containers/case-home/case-home.component.ts
```diff
@@ -108,7 +114,8 @@ export class CaseHomeComponent implements OnInit, OnDestroy {
     if (error.status !== 401 && error.status !== 403) {
       console.log('error during triggering event:', triggerId);
```
paddybasi: trivial, but should we not be using a logger rather than console.log

PR #367, opened 2020-02-11
EUI-1349 updated package.json

adnanakgun:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -60,8 +60,8 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "2mb",
-                  "maximumError": "5mb"
+                  "maximumWarning": "6mb",
+                  "maximumError": "8mb"
                 }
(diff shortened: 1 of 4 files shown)


angular.json
```diff
@@ -60,8 +60,8 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "2mb",
-                  "maximumError": "5mb"
+                  "maximumWarning": "6mb",
```
ronaldmansveld: Why are these limits raised so much? What has changed that the app is now so big (and therefore heavy to load)? Have we looked into ways to make the app smaller?

adnanakgun: We had a look at this with @paddybasi on master before and had decided to defer this. Basically, this is the same as master. Maybe we should create a separate ticket to deal with this.

PR #514, opened 2020-05-28
Feature/eui 1882 session timeouts

ficklephil:
Adding the Session Timeout Modal.
Session Timeouts calculated in the Node layer.
Use of different Session Timeouts for different User on Manage Cases, set via configuration.

For a selection of timeout times please see the following Jira ticket.
https://tools.hmcts.net/jira/browse/EUI-1882


src/app/containers/app/app.component.ts
```diff
@@ -1,16 +1,191 @@
...
+   *  "pattern": "-solicitor",
+   *  "totalIdleTime": 50
+   *  }
+   * }
+   */
+  public userDetailsHandler(userDetails) {
+
+    if (propsExist(userDetails,['sessionTimeout']) && userDetails.sessionTimeout.totalIdleTime > 0) {
+
+      console.log(userDetails.sessionTimeout);
```
ficklephil, line 67: Temporary, making it easier for QA to test.

api/user/index.ts
```diff
@@ -0,0 +1,29 @@
...
+interface SessionRequest extends express.Request {
+  session: {
+    user: {
+      roles: ['']
+    }
+  },
+}
+
+// TODO: Need to test error condition here.
+export async function getUserDetails(req: SessionRequest, res: express.Response, next: express.NextFunction) {
```
paddybasi, line 15: could we not do this please? I never liked the enhancedRequest etc before either. Instead try doing a declaration merging, but only if the @types aren't available

src/app/services/auth/auth.service.ts
```diff
@@ -81,8 +81,20 @@ export class AuthService {
...
     window.location.href = '/api/logout';
   }
+
+  logOutAndRedirect() {
+
+    console.log('logOutAndRedirect');
+    const { hostname, port } = window.location;
+    const portNumber = port ? `:${port}` : '';
+
+    const baseUrl = `http://${hostname}${portNumber}`;
```
paddybasi, line 95: please don't do this, let the backend do it? also protocol will be wrong etc, not always http

api/user/index.ts
ficklephil: Fixed after call with Paddy.

src/app/services/auth/auth.service.ts
ficklephil: Awaiting the Open Id logout route, then we will refactor this as discussed with Paddy.

PR #519, opened 2020-05-28
error handling within interceptors.ts

paddybasi:

### Change description ###

```
[ ] Yes
[ ] No
```


api/lib/interceptors.ts
```diff
@@ -54,6 +54,9 @@ export function errorInterceptor(error) {
     let data = valueOrNull(error, 'response.data.details')
     if (!data) {
         data = valueOrNull(error, 'response.status') ? JSON.stringify(error.response.data, null, 2) : null
+        if (!data) {
+            data = error
+        }
         logger.error(`Error on ${error.config.method.toUpperCase()} to ${url} in (${error.duration}) - ${error} \n
```
timja, line 60: is it guaranteed that `error.config.method` will always be set?

paddybasi: from what I can see yes, have done a debug session (changed the ccd url etc). This comes from axios directly

timja: are all errors going to be from API calls though? it's probably good enough for now, but possibly needs re-visiting later

paddybasi: this interceptor will only get called for any failed axios requests so hopefully good enough for now. But agreed it won't catch any TypeErrors etc

PR #527, opened 2020-06-03
Remove log token eui-2066

borislavpetrovhmcts:
Remove log token
Resolves eui-2066

### DIFF
--- a/api/lib/error.handler.ts
+++ b/api/lib/error.handler.ts
@@ -13,9 +13,9 @@ export default function errorHandler(err, req: Request, res: Response, next: Nex
         // remove any sensitive data, such as bearer token from being logged
         delete err.config.headers
     }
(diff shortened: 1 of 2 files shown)


api/lib/error.handler.ts
```diff
@@ -13,6 +13,10 @@ export default function errorHandler(err, req: Request, res: Response, next: Nex
         // remove any sensitive data, such as bearer token from being logged
         delete err.config.headers
     }
+    if (propsExist(err, ['request', '_header'])) {
```
adnanakgun, line 16: Can you test that as well please?

borislavpetrovhmcts: There are no tests for this file. Plus this is urgent

borislavpetrovhmcts: Unit test added

PR #536, opened 2020-06-15
Bug/eui 2102 retrieving past query aa

adnanakgun:
# JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/EUI-2102

### Change description ###

Added check for jurisdictions and fixed tests

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/cases/containers/case-list/case-list.component.ts
+++ b/src/cases/containers/case-list/case-list.component.ts
@@ -128,23 +128,25 @@ export class CaseListComponent implements OnInit, OnDestroy {
   public setCaseListFilterDefaults = () => {
     this.jurisdictionsBehaviourSubject$
       .subscribe(jurisdictions => {
-        this.savedQueryParams = JSON.parse(localStorage.getItem('savedQueryParams'));
-        if (this.savedQueryParams && this.savedQueryParams.jurisdiction && !this.doesIdExist(this.jurisdictions, this.savedQueryParams.jurisdiction)) {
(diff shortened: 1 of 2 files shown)


src/cases/containers/case-list/case-list.component.ts
```diff
@@ -128,23 +128,25 @@ export class CaseListComponent implements OnInit, OnDestroy {
...
+          if (this.savedQueryParams) {
             this.defaults = {
-              jurisdiction_id: jurisdictions[0].id,
-              case_type_id: jurisdictions[0].caseTypes[0].id,
-              state_id: jurisdictions[0].caseTypes[0].states[0].id
+              jurisdiction_id: this.savedQueryParams.jurisdiction,
+              case_type_id: this.savedQueryParams['case-type'],
+              state_id: this.savedQueryParams['case-state']
             };
+          } else {
```
paddybasi, line 142: maybe remove the trailing ```{``` and have it as ```else if { ```

adnanakgun:
> please include following scenario as part of unit tests.
> 
> Should set defaults from users jurisdictions when "savedQueryParams" jurisdiction is not present in users jurisdictions
> 
> There was a bug around this scenario in past, https://tools.hmcts.net/jira/browse/EUI-1514 .

I believe this case to be covered already in this test:
`setCaseListFilterDefaults() should set the defaults.` line 318 in case-list.component.spec.ts

However I have added an additional assertion to be extra sure.

PR #547, opened 2020-06-24
Feature/eui 1948 caseshare by type aa

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-1948

### Change description ###

Utilised LD

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/cases/containers/case-list/case-list.component.spec.ts
+++ b/src/cases/containers/case-list/case-list.component.spec.ts
@@ -319,6 +319,42 @@ describe('CaseListComponent', () => {
     });
   });
 
+  describe('isCaseShareVisible()', () => {
+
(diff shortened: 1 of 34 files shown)


src/cases/containers/case-list/case-list.component.spec.ts
```diff
@@ -312,6 +319,42 @@ describe('CaseListComponent', () => {
     });
   });
 
+  describe('isCaseShareVisible()', () => {
```
sreekanthpuligadda, line 322: Please add following test for input values.
isCaseShareVisible(true, [])
isCaseShareVisible(false, ['dummy'])
isCaseShareVisible(false, [])

PR #551, opened 2020-06-29
Feature/share-a-case-eui-1984-1996

rayliangatos:

### Change description ###

```
[ ] Yes
[x ] No
```


api/tslint.json
```diff
@@ -46,7 +46,7 @@
         ],
         "quotemark": false,
         "semicolon": [
-            true,
+            false,
```
adnanakgun, line 49: Don't change this please. I'm not a fan either, but the whole node layer is written with non-semicolon style.

api/caseshare/stub-api.ts
```diff
@@ -0,0 +1,95 @@
...
+    const searchText = req.query.q;
+    const org = getOrgById(req.params.orgId);
+    if (!org) {
+      return res.status(404).send('{"errorMessage": "Organisation is not found}"');
+    }
+    if (searchText === undefined) {
+      return res.send(org.users);
+    } else {
+      const users = org.users;
+      const user = users.filter(u => u.idamId.toLowerCase().indexOf(searchText.toLowerCase()) > -1 ||
```
ficklephil, line 64: It's hard to know what this filter is doing just by looking at it. It's worth abstracting this filter code out into a separate filter func, and passing the named filter function into here.

config/default.json
```diff
@@ -23,6 +23,9 @@
     "documents": {
       "api": "http://dm-store-prod.service.core-compute-prod.internal"
     },
+    "caseShare": {
+      "api": "http://ccd-case-share-api-aat.service.core-compute-aat.internal"
```
adnanakgun, line 27: This should be prod. Use local configuration for aat.

api/caseshare/stub-api.ts
```diff
@@ -0,0 +1,95 @@
...
+      return res.status(404).send('{"errorMessage": "Cases are not found}"');
+    }
+    return res.send(cases);
+}
+
+export function getCaseById(req: EnhancedRequest, res: Response) {
+    if (!cases) {
+      return res.status(404).send('{"errorMessage": "Cases are not found}"');
+    }
+    const foundCase = cases.find(c => c.caseId === req.params.caseId);
```
ficklephil, line 86: Move away from using single letters as variable names. As with larger functions it makes it difficult to understand what's going within a func.

src/cases/store/reducers/share-case.reducer.ts
```diff
@@ -1,32 +1,78 @@
...
+      };
+    case ShareCasesActions.LOAD_SHARE_CASES_FAILURE:
+      return {
+        ...state,
+        error: action.payload,
+        loading: false
+      };
+    case ShareCasesActions.ADD_SHARE_CASES:
+      const addShareCases = state.shareCases.slice();
+      for (let i = 0, l = action.payload.sharedCases.length; i < l; i++) {
```
ficklephil, line 52: const

src/cases/models/organisation/organisation.module.ts
```diff
@@ -0,0 +1,7 @@
+import { UserDetails } from '../user-details/user-details.module';
+
+export interface OrganisationModule {
```
adnanakgun, line 3: Module or Model? I think the filename as well as the interface name for UserDetail is also wrong.

proxy.config.json
```diff
@@ -9,6 +9,11 @@
     "secure": false,
     "logLevel": "debug"
   },
+  "/api/caseshare/*": {
```
paddybasi, line 12: This isn't required as the route below will catch it

api/caseshare/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// TODO: use below url for actual API
+const url: string = getConfigValue(SERVICES_CASE_SHARE_API_URL);
+
+/**
+ * getRoot
+ * example: /api/caseshare/
+ */
+export async function getRoot(req: EnhancedRequest, res: Response) {
+  if (process.env.STUB) {
```
paddybasi, line 15: Please don't use process.env, instead look at how we have the ```config``` folder

api/caseshare/stub-api.ts
```diff
@@ -0,0 +1,95 @@
+import { plainToClass } from 'class-transformer';
+import { Response } from 'express';
+import { SharedCase } from '../../src/cases/models/case-share/case-share.module';
```
paddybasi, line 3: Please don't mix models between api layer and angular. We've had issues with this in the past

api/caseshare/stub-api.ts
rayliangatos: comment addressed on new PR:
https://github.com/hmcts/rpx-xui-webapp/pull/553/files

PR #567, opened 2020-07-07
Eature/eui 2187 integrate xui node and common lib

paddybasi:

### Change description ###

```
[ ] Yes
[ ] No
```


api/application.ts
```diff
@@ -1,30 +1,26 @@
 import * as bodyParser from 'body-parser'
 import * as cookieParser from 'cookie-parser'
 import * as express from 'express'
-import * as session from 'express-session'
 import * as helmet from 'helmet'
 // import {router as termsAndCRoutes} from './termsAndConditions/routes'
```
rayliangatos, line 5: though this is not part of the changes, can these unuse imports be tidied up

api/auth/index.ts
```diff
@@ -1,69 +1,130 @@
...
+                s2sEndpointUrl: `${getConfigValue(SERVICE_S2S_PATH)}/lease`,
+                s2sSecret: s2sSecret.trim(),
+            },
+        },
+        session: showFeature(FEATURE_REDIS_ENABLED) ? redisStoreOptions : fileStoreOptions,
+    }
+
+    const type = showFeature(FEATURE_OIDC_ENABLED) ? 'oidc' : 'oauth2'
+    nodeLibOptions.auth[type] = options
+    logger._logger.info('Setting XuiNodeLib options')
```
rayliangatos, line 128: can this logger._logger be simplified?

api/lib/proxy.ts
```diff
@@ -12,38 +12,43 @@ import {
...
 
 import { http } from './http'
-//import * as log4jui from './log4jui'
 import { EnhancedRequest } from './models'
+import {exists} from './util'
 
-//const logger = log4jui.getLogger('proxy')
-
-export function setHeaders(req: EnhancedRequest) {
+export function setHeaders(req: express.Request) {
```
rayliangatos, line 18: can this be simplified to Request

api/lib/proxy.ts
```diff
@@ -12,38 +12,43 @@ import {
...
+        if (exists(req, 'headers.ServiceAuthorization')) {
+            headers.ServiceAuthorization = req.headers.ServiceAuthorization
+        }
     }
 
     return headers
 }
 
-export async function get(req: EnhancedRequest, res: express.Response, next: express.NextFunction) {
+export async function get(req: express.Request, res: express.Response, next: express.NextFunction) {
```
rayliangatos, line 51: can this be simplified to Request, Response, NextFunction?

src/app/services/auth/auth.service.ts
```diff
@@ -1,106 +1,38 @@
...
+  public logOut(): Observable<any> {
+    return this.httpService.get('/auth/logout?noredirect=true');
   }
 
-  public getWindowLocation() {
-    return window.location;
+  public logOutAndRedirect() {
+    this.logOut().subscribe( () => {
+      this.setWindowLocationHref('/idle-sign-out')
+    })
```
rayliangatos, line 35: missing ; which has tslint issue

api/user/index.ts
```diff
@@ -5,7 +5,7 @@ import { getUserSessionTimeout } from './userTimeout'
 
 export async function getUserDetails(req, res: Response, next: NextFunction) {
 
-  const { roles } = req.session.user
+  const { roles } = req.session.passport.user.userinfo
```
rayliangatos, line 8: this looks awkard any reason to wrap with passport

api/application.ts
paddybasi: we intend to put T&C's back in once it's been deployed to production

api/auth/index.ts
paddybasi: This is how the 'old' logger was coded, exposing the underlying logger as _logger (I prefer to use this as it's not destructive, the old one would convert objects to [Object Object])

api/lib/proxy.ts
paddybasi: Ideally we need a clean up task of the node layer, including removing of ```EnhancedRequest``` however this PR isn't the place for it :)

api/lib/proxy.ts
paddybasi: Yes it can, again a clean up task (tech debt for later)

api/user/index.ts
paddybasi: the library we've pushed uses passport under the hood; if we then stored user back onto ```req.session``` there would be a lot of data that would have to be transferred back/forth to redis.

src/app/services/auth/auth.service.ts
paddybasi: fixed, though it would only have given a warning (which we have plenty of anyway :joy: )

PR #580, opened 2020-07-21
Feature/eui 1999 case share complete

kpokala:

### Change description ###

```
[ ] Yes
[ ] No
```


src/cases/store/effects/share-case.effects.ts
```diff
@@ -72,4 +74,20 @@ export class ShareCaseEffects {
...
   );
+
+  @Effect()
+  public assignUsersWithCases1$ = this.actions$.pipe(
+    ofType(shareCaseActions.ASSIGN_USERS_TO_CASE),
+    map((action: shareCaseActions.AssignUsersToCase) => action.payload),
+    switchMap(payload => {
+      this.payload = payload;
+      const ret: Observable<SharedCase[]> = this.caseShareService.assignUsersWithCases(payload);
+      return ret.pipe(
```
udaydenduluri33, line 85: not using the reference ret further down. So, can club both the statments and get rid of ret

src/cases/containers/case-share-complete/case-share-complete.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.newShareCases$ = this.store.pipe(select(fromCasesFeature.getShareCaseListState));
+    this.newShareCases$.subscribe(shareCases => {
+      this.completeScreenMode = this.checkIfIncomplete(shareCases)
+      this.newShareCases = shareCases
+    });
+
+  }
+
+  private checkIfIncomplete(shareCases: SharedCase[]) {
+    for (const aCase of shareCases) {
```
udaydenduluri33, line 40: Try using some and also, could add some unit tests for it

src/cases/store/effects/share-case.effects.ts
kpokala: Defined a separate one to print console.log, forgot to remove it..resilved now

api/caseshare/index.ts
```diff
@@ -1,7 +1,14 @@
+import {SharedCase} from '@hmcts/rpx-xui-common-lib/lib/models/case-share.model'
+import * as express from 'express'
 import { Response } from 'express'
+import * as fromCasesFeature from '../../src/cases/store';
```
paddybasi, line 4: This will fail the build

api/caseshare/stub-api.ts
```diff
@@ -2,14 +2,14 @@ import { SharedCase } from '@hmcts/rpx-xui-common-lib/lib/models/case-share.mode
 import { UserDetails } from '@hmcts/rpx-xui-common-lib/lib/models/user-details.model'
 import { plainToClass } from 'class-transformer'
 import { Response } from 'express'
-import { OrganisationModel } from '../../src/cases/models/organisation/organisation.model'
+//import { OrganisationModel } from '../../src/cases/models/organisation/organisation.model'
 import { EnhancedRequest } from '../lib/models'
 import { DataBaseModel } from './models/data-base.model'
 // @ts-ignore
 import * as dbJson from './stubs/db.json'
```
paddybasi, line 9: Not sure if we should be supporting imports of json (yes it does work though)

api/caseshare/index.ts
```diff
@@ -1,7 +1,14 @@
+import {SharedCase} from '@hmcts/rpx-xui-common-lib/lib/models/case-share.model'
```
paddybasi, line 1: common-lib is a frontend, while the model is just purely a js object think the preference would be for api to have it's own models

tsconfig.json
```diff
@@ -9,12 +9,14 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "allowJs": true,
```
paddybasi, line 12: can we remove this please, we don't want to be hacking our tsconfig to get it working with normal js code as well

PR #623, opened 2020-08-25
Apply fixes for page title

ronaldmansveld:

### Change description ###

```
[ ] Yes
[ ] No
```


src/app/containers/app/app.component.ts
```diff
@@ -17,15 +19,30 @@ export class AppComponent implements OnInit {
...
-    private readonly timeoutNotificationsService: TimeoutNotificationsService
+    private readonly googleAnalyticsService: GoogleAnalyticsService,
+    private readonly timeoutNotificationsService: TimeoutNotificationsService,
+    private readonly router: Router,
+    private readonly titleService: Title
   ) {
 
     this.googleAnalyticsService.init(config.googleAnalyticsKey);
+
+    this.router.events.subscribe((data) => {
```
adnanakgun, line 34: Maybe assign to Subscription so you can unsubscribe in onDestroy?

ronaldmansveld: No need as this is the appcomponent, which lives throughout the whole application livecycle. If it was a deeper component yes, but here it's not necessary, so better from a memory point of view to not keep that reference around.

PR #637, opened 2020-09-17
Feature/eui 2292 judicial header 2

ficklephil:
The header is now dynamic and style is dependent on the Users role.
The header is now easily adjustable, with properties being in the app.constants.ts file.
The header is now easily adjustable for different User groups.

A Judicial User is able to log in ie. juitestjudgesscs@mailnesia.com
and see the Judicial Header Styling and Navigation (A Judicial User only has 'Case list' as their navigation item).

A Case Manager User is able to log in ie. lukesuperuserxui@mailnesia.com
and see the MyHMCTS Header styling.

A default user is able to log in ie. probatebackoffice@gmail.com and see the default Header and styling.

On the Terms and Conditions page, there is still no navigation, as before.
Sign out works for all Users.
Idle sign out page is available for all users.


udaydenduluri33 (approved):
LGTMp

PR #642, opened 2020-09-22
upgraded to latest node lib - csrf feature

paddybasi:
https://tools.hmcts.net/jira/browse/EUI-1957

### Change description ###
upgrade node lib to implement csrf feature

```
[ ] Yes
[x] No
```


api/auth/index.ts
```diff
@@ -67,7 +67,7 @@ export const getXuiNodeMiddleware = () => {
         callbackURL: getConfigValue(SERVICES_IDAM_OAUTH_CALLBACK_URL),
         clientID: idamClient,
         clientSecret: secret,
-        discoveryEndpoint: `${idamWebUrl}/o`,
+        discoveryEndpoint: `${idamWebUrl}/o/.well-known/openid-configuration`,
```
ronaldmansveld, line 70: directories that start with a . are generally hidden/system dirs. Are we sure we are supposed to hard code it this way?

Mo-Lala84 (approved):
lgtm

api/auth/index.ts
paddybasi: it's not a directory, but the standard for autodiscovery endpoints @see https://ldapwiki.com/wiki/Openid-configuration#:~:text=Openid%2Dconfiguration%20is%20a%20Well,the%20Identity%20Provider%20(IDP).

paddybasi: However I do agree that perhaps it shouldn't be hardcoded, maybe a task in future to move this into the config instead

Mo-Lala84: its inline with guidelines, not a show stopper but a good call to query it.

ronaldmansveld (approved):

PR #649, opened 2020-09-25
Fix aria on filter button

ronaldmansveld:

### Change description ###

```
[ ] Yes
[ ] No
```


src/cases/containers/case-list/case-list.component.ts
```diff
@@ -209,6 +209,7 @@ export class CaseListComponent implements OnInit, OnDestroy {
       this.paginationMetadata.total_results_count = paginationMetadata.total_results_count;
 
       const event = this.getEvent();
+      // tslint:disable-next-line: triple-equals
```
paddybasi, line 212: Is there any particular reason for not using ```===```?

paddybasi (approved):

src/cases/containers/case-list/case-list.component.ts
ronaldmansveld: in this case yes: != checks against null and undefined, !== would only check against null, that's why I added the disable rule comment, as we want that behaviour here.

paddybasi: okay no worries, perhaps add that as a comment for the next unsuspecting developer :sweat_smile:

PR #681, opened 2020-10-26
Feature/eui 2292 judicial header 5

ficklephil:
Judicial Header.


Mo-Lala84 (approved):
approved

yarn.lock
```diff
@@ -1569,6 +1577,14 @@ amdefine@>=0.0.4:
   resolved "https://registry.yarnpkg.com/amdefine/-/amdefine-1.0.1.tgz#4a5282ac164729e93619bcfd3ad151f817ce91f5"
   integrity sha1-SlKCrBZHKek2Gbz9OtFR+BfOkfU=
 
+angular-tree-component@^8.5.6:
```
adnanakgun, line 1580: Has the yarn.lock file been deleted and rebuild in this PR? Why are there so many changes to the lock file?

adnanakgun (approved):

src/app/app.constants.ts
```diff
@@ -30,29 +31,109 @@ const FooterDataNavigation = {
...
+    roles: [
+      'caseworker-sscs-judge',
+      'caseworker-sscs-panelmember',
+      'caseworker-cmc-judge',
+      'caseworker-divorce-judge',
+      'caseworker-divorce-financialremedy-judiciary',
+      'caseworker-probate-judge',
+      'caseworker-ia-iacjudge',
+      'caseworker-publiclaw-judiciary',
+    ],
```
rayliangatos, line 45: roles hard coded? possible to move to LD or config?

ficklephil: It is in the LD Config, this is the fall over, just in case LD stops working.

rayliangatos (approved):

PR #695, opened 2020-11-09
Task Home container with Global Header and Footer.

ficklephil:
Task Home container with Global Header and Footer.


udaydenduluri33 (approved):
LGTMp

paul-graham (changes requested):
As discussed, the `work-allocation-feature.routes.ts` file needs to be adjusted to make the task list and task manager children so they show in the `<router-outlet></router-outlet>`.

src/work-allocation/work-allocation-feature.routes.ts
```diff
@@ -1,14 +1,18 @@
...
 
 export const ROUTES: Routes = [
     {
       path: '',
-      component: TaskListComponent
+      component: TaskHomeComponent
     },
+    // TODO: We probably don't need the following routes,
+    // as these components parents should be TaskHomeComponent.
+    // to discuss.
```
paul-graham, line 15: `src/cases/case-feature.routes.ts` shows an example of this type of routing. I think the URL will still need to be distinct to differentiate between viewing the `TaskListComponent` and the `TaskManagerComponent`.

My suggestion would be to make the `task-list` route just `list` and the `task-manager` route just `manager` and then make both routes children of the default route (`path: ''`) so we get urls of `/tasks/list` and `/tasks/manager`.

paul-graham (changes requested):

PR #705, opened 2020-11-11
Feature/workallocation/eui 2845 task list component

ficklephil:
Task List Component the Controller and 30% of the view are now complete.

Unit tests need to be written but will be handed over so that I can pick up 2844.

PR with code: [https://github.com/hmcts/rpx-xui-webapp/pull/705]

I've followed the LDD closely, therefore the following is the copy from the LLD, with if this functionality has been covered by the TaskListComponent.
 * The main input for this component is an array of TaskModels. These tasks must be passed in the order they will be displayed, no ordering will happen within the task list component itself. - *_DONE_***
 * In order to display the correct columns for tasks, the component will contain an input for a TaskServiceConfig, as defined in the HLD section for UI Config, as well as an input to define the desired view using the TaskView enum. - The component has an input for fields which is part of the TaskFieldConfig[] object,
thoughts around this were, the sortDirection is handled by the parent, I wanted to try and keep this component generic therefore I'm only passing in fields, as (I think) it's the only property of the TaskServiceConfig that would be used within this TaskListComponent. (to Discuss)
 * The component will contain an Output for actions that are triggered on a task, modelled using the InvokedTaskAction model. - *_DONE_*
as well as an output to trigger sorting of the displayed data. - *_DONE_*
 * Pseudocode adhered too - *_DONE_*

Things to note in the pseudocode:
 * The dataSource is an Observable of the data to be displayed fields is the property of TaskFieldConfig[], containing the configuration for the fields as returned by the API - *_DONE_*
 * sort() is a function that takes in the fieldname, so it can be outputted to trigger a new request to the API to get a sorted resultset - *_DONE_*
 * formatValue() is a function that takes in the value and the fieldType to properly render the value. - We're not using this anymore / different ticket.
 * displayedColumns is an array of strings specifying the fieldnames to be displayed, in order, this will have to include the 'manage' column as the last one. - *_DONE_*

Many thanks,
Phil


connorpgpmcelroy (approved):
Looks good to me

src/work-allocation/containers/task-list/task-list.component.ts
```diff
@@ -1,10 +1,157 @@
...
+   */
+  public dataSource$: Observable<Task[]>;
+
+  public displayedColumns: string[];
+
+  private selectedRow: Task;
+
+  public ngOnChanges() {
+    if (this.dataSource$) {
+      this.dataSource$.subscribe().unsubscribe();
```
udaydenduluri33, line 43: typo?

paul-graham: Mere stupidity. :)

udaydenduluri33 (approved):
LGTMp

PR #721, opened 2020-11-13
EUI-2921 - Case reference field

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2921

### Change description ###
Case reference field, primarily for use in the Task list.

```
[ ] Yes
[x] No
```


connorpgpmcelroy (approved):
Looks good to me

connorpgpmcelroy (approved):

src/work-allocation/components/case-reference-field/case-reference-field.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  @Input() public caseReference: string;
+
+  private pHref: string;
+  public ngOnChanges(): void {
+    let href;
+    if (this.caseReference) {
+      // Get rid of the spaces in the caseReference.
+      const caseId = this.caseReference.replace(/\s/g, '');
+      if (caseId) {
+        href = `/cases/case-details/${caseId}`;
```
udaydenduluri33, line 20: This could be a global constant

src/work-allocation/components/case-reference-field/case-reference-field.component.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    }));
+
+    beforeEach(() => {
+      fixture = TestBed.createComponent(WrapperComponent);
+      wrapper = fixture.componentInstance;
+      component = wrapper.appComponentRef;
+      fixture.detectChanges();
+    });
+
+    it('should handle a CASE_REFERENCE type', () => {
```
ronaldmansveld, line 37: Is there a reason all the checks are in one single test instead of separate tests?

src/work-allocation/components/case-reference-field/case-reference-field.component.ts
paul-graham: It could, and I looked for one, but I didn't find a global constant anywhere. It's inline like that throughout the codebase. If it's introduced, I think it would make sense to use it for all `/cases/case-details/` instances.

src/work-allocation/components/task-field/task-field.component.spec.ts
```diff
@@ -617,6 +617,58 @@ describe('WorkAllocation', () => {
       expect(fixture.debugElement.nativeElement.querySelector('img')).toBeNull();
     });
 
+    it('should handle a CASE_REFERENCE type', () => {
```
ronaldmansveld, line 620: As above, why not multiple tests?

ronaldmansveld (changes requested):

src/work-allocation/components/case-reference-field/case-reference-field.component.ts
ronaldmansveld: Agree with Uday, the base part of that URL should not be stored in the component, but come from config.

ficklephil: Will '/cases/case-details/' change periodically? If yes then the configuration is the correct place for it. If no then it's better as a constant within the application. Why? - Because if we placed all strings into our configuration, we would end up with bloat at two points: 1. Within the .json configuration files, and 2. within the configuration/reference file.

ficklephil (approved):

src/work-allocation/components/case-reference-field/case-reference-field.component.spec.ts
paul-graham: Fixed.

src/work-allocation/components/task-field/task-field.component.spec.ts
paul-graham: Fixed.

src/work-allocation/components/case-reference-field/case-reference-field.component.ts
paul-graham: I've moved it to `AppConstants`. If we need it to be configurable at some point, that's likely going to need an application-wide change.

udaydenduluri33 (approved):
LGTMp

ronaldmansveld (approved):

PR #733, opened 2020-11-18
Fixed EUI-2936

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2936

### Change description ###
Fixed a bug with a binding in the Case list component.

```
[ ] Yes
[x] No
```


Daniel-Lam (approved):
Approved, @paul-graham

src/cases/containers/case-list/case-list.component.ts
```diff
@@ -94,6 +112,11 @@ export class CaseListComponent implements OnInit, OnDestroy {
       this.jurisdictions = jurisdictions;
     });
 
+    this.featureToggleSubscription = this.featureToggleService.getValue('shareable-jurisdictions', []).subscribe(value => {
+      this.shareableJurisdictions = value;
+      this.setupCaseShareVisibility();
```
rayliangatos, line 117: Agreed that the two aysch pipes cause the errors, but the 'this.setupCaseShareVisibility();' and another one in line 156 would probably cause race conditions when we put this trigger(this.setupCaseShareVisibility) in two subscription blocks. Ideally we need to merge or switch the two observables into one isCaseShareVisible$ and still use the asynch pipe like (isCaseShareVisible$ | asynch) in the html.

paul-graham: I don't follow why it would cause race conditions. The `setupCaseShareVisibility()` method checks for the presence of data from both subscriptions before proceeding. I'm not entirely sure what would be gained from putting an async pipe in the HTML...?

ronaldmansveld (approved):

ronaldmansveld (approved):

paul-graham:
Closed by the bot because it was "stale", but it's still needed.

paul-graham:
Ended up merged into https://github.com/hmcts/rpx-xui-webapp/pull/760 for some reason.

PR #735, opened 2020-11-19
EUI-2869: task service (node)

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2869

### Change description ###
Added a bunch of unit tests for the task service in the node layer. Also a small bit of refactoring to better name some things.

```
[ ] Yes
[x] No
```

### DIFF
--- a/api/test/pact/pact-tests/work-allocation-post-task.spec.ts
+++ b/api/test/pact/pact-tests/work-allocation-post-task.spec.ts
@@ -1,11 +1,11 @@
-import { Pact } from '@pact-foundation/pact'
-import { assert, expect } from 'chai'
-import * as getPort from 'get-port'
-import * as path from 'path'
+import { Pact } from '@pact-foundation/pact';
+import { assert, expect } from 'chai';
+import * as getPort from 'get-port';
+import * as path from 'path';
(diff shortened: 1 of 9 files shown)


connorpgpmcelroy (approved):

connorpgpmcelroy (commented):
Might be a silly question but why are the changes just semi-colon removals for the imports on this commit? Is this necessary?

paul-graham:
> Might be a silly question but why are the changes just semi-colon removals for the imports on this commit? Is this necessary?

Not a silly question at all... but it's a silly reason if you ask me. Basically, semicolons are allowed in the Angular layer but they are _forbidden_ in the node layer. I think they should simply be allowed (and even encouraged) in both.

What happened here is that when I use the command to organise imports in my IDE (Visual Studio Code) it does the organising nicely but it also adds semicolons to the end of the lines because, foolishly, it thinks I actually _want_ fully valid TypeScript. :)

connorpgpmcelroy (approved):

connorpgpmcelroy (approved):

udaydenduluri33 (approved):
LGTMp

PR #737, opened 2020-11-19
Feature/eui 2871 create caseworker

udaydenduluri33:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
+++ b/api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
@@ -1,0 +1,67 @@
+import { Pact } from '@pact-foundation/pact'
+import { assert } from 'chai'
+import * as getPort from 'get-port'
+import * as path from 'path'
+import { EnhancedRequest } from '../../../lib/models'
+import { handleCaseWorkerDetails } from '../../../workAllocation/caseWorkerService'
+
+describe("Work Allocation for location and service Caseworker API", () => {
+
+    let MOCK_SERVER_PORT
+    let workallocationUrl
+    let provider
+    const mockResponse = {
+          "firstName": "firstName",
+          "lastName": "lastName",
+          "idamId": "XXX-XXX-XX",
+          "location": {
+            "id": "string",
+            "locationName": "string",
+            "services": []
+          },
+        }
+
+    before(async () => {
+        MOCK_SERVER_PORT = await getPort()
+        workallocationUrl = `http://localhost:${MOCK_SERVER_PORT}`
+        provider = new Pact({
+          consumer: 'xui_get_caseworker_caseworkerId',
+          dir: path.resolve(__dirname, '../pacts'),
+          log: path.resolve(__dirname, '../logs', 'work-allocation.log'),
+          logLevel: 'info',
+          port: MOCK_SERVER_PORT,
+          provider: 'WorkAllocation_api',
+          spec: 2,
+        })
+        return provider.setup()
+      })
+
+    // Write Pact when all tests done
+    after(() => provider.finalize())
+
+      // verify with Pact, and reset expectations
+    // afterEach(() => provider.verify())
+
+    describe('when a request to get caseworker for id', () => {
+        before(() =>
+          provider.addInteraction({
+            state: '.well-known endpoint',
+            uponReceiving: 'a request for configuration',
+            withRequest: {
+              method: 'GET',
+              path: '/caseworker/caseworker123',
+            },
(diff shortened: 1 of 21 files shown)


paul-graham (changes requested):
All the get methods in `caseWorkerService.ts` are essentially identical. The only difference is with what gets logged. It should be refactored to modularise everything _other_ than the logging or to have a single get method with a more generic log.

I'd favour the single get method as it already logs the path, which I would argue is sufficient distinction between the different endpoints.

api/test/pact/pact-tests/get-for-location-caseworker.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { Pact } from '@pact-foundation/pact'
+import { assert } from 'chai'
+import * as getPort from 'get-port'
+import * as path from 'path'
+import { EnhancedRequest } from '../../../lib/models'
+import { handleCaseWorkerForLocation } from '../../../workAllocation/caseWorkerService'
+
+describe("Work Allocation for location Caseworker API", () => {
+
+    let MOCK_SERVER_PORT
```
ficklephil, line 10: Not sure Capitals should be used here, as this is not a static constant.

api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { Pact } from '@pact-foundation/pact'
+import { assert } from 'chai'
+import * as getPort from 'get-port'
+import * as path from 'path'
+import { EnhancedRequest } from '../../../lib/models'
+import { handleCaseWorkerDetails } from '../../../workAllocation/caseWorkerService'
+
+describe("Work Allocation for location and service Caseworker API", () => {
```
ficklephil, line 8: We're using both single quotes and double quotes within this file, let's just use single quotes for consistency.

api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    let MOCK_SERVER_PORT
+    let workallocationUrl
+    let provider
+    const mockResponse = {
+          "firstName": "firstName",
```
ficklephil, line 14: Double quotes on the name, not required here.

api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
paul-graham: Good spot!

udaydenduluri33:
> All the get methods in `caseWorkerService.ts` are essentially identical. The only difference is with what gets logged. It should be refactored to modularise everything _other_ than the logging or to have a single get method with a more generic log.

We do not have the Swagger so I will leave it like this. Having said the point is noted. Agree in Principle. @paul-graham

api/test/pact/pact-tests/get-for-location-caseworker.spec.ts
udaydenduluri33: This is the normal convention for all the PACT tests. Happy to change it. but that means 32 changes in 8 files. Are you sure about this?

api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
udaydenduluri33: Done

paul-graham:
> We do not have the Swagger so I will leave it like this. Having said the point is noted. Agree in Principle. @paul-graham

Isn't that _more_ the reason we should make it super simple now? We only need to complicate matters if it turns out that the Swagger isn't as simple as it seems...?

api/test/pact/pact-tests/get-caseworker-by-id.spec.ts
udaydenduluri33: Done

paul-graham (approved):
Approved, with an understanding that my comments will be taken into account during the integration phase and duplicate code _must_ be sorted.

connorpgpmcelroy (approved):
I can probably derive some of this for the location-data-service

api/test/pact/pact-tests/get-for-location-caseworker.spec.ts
ficklephil: Let's change it as we go :) It's a development convention that static constants are in full capitals. But nothing else should be.

ficklephil (approved):

PR #747, opened 2020-11-24
Semicolons

paul-graham:
### Change description ###
Allow semicolons in the node layer to align it with the Angular layer. It will now warn about missing semicolons.

```
[ ] Yes
[x] No
```


ficklephil (approved):

Mo-Lala84 (approved):
approved

rayliangatos (approved):
Do you need another PR to add the ; back to every file to make the codes be consistent?

paul-graham:
> Do you need another PR to add the ; back to every file to make the codes be consistent?

If we're going to do that, we will. For now, though, this simply allows them again and warns if they're not in place, which is exactly what we have in the node layer.

paul-graham:
Closed by the bot because it was "stale", but it's still needed.

paul-graham:
Appears to have been merged by some other means.

PR #773, opened 2020-12-03
Feature/eui 2897 wa inline messaging

ficklephil:
Information, Success and Warning visuals.

### DIFF
--- a/src/work-allocation/components/info-message/info-message.component.ts
+++ b/src/work-allocation/components/info-message/info-message.component.ts
@@ -1,36 +1,24 @@
-import {Component, Input, OnInit, ViewEncapsulation} from '@angular/core';
+import {Component, Input, OnInit} from '@angular/core';
 
-import { TaskFieldType } from './../../enums';
-import { Task, TaskFieldConfig } from './../../models/tasks';
+import {InfoMessage, InfoMessageType} from './../../enums';
 
 @Component({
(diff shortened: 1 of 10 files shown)


src/work-allocation/components/info-message/info-message.component.ts
```diff
@@ -0,0 +1,24 @@
...
+   */
+  @Input() public infoMessageType: InfoMessageType;
+
+  @Input() public infoMessage: InfoMessage;
+
+  public infoMessageType = InfoMessageType;
+
+  constructor() {}
+
+  public ngOnInit(): void {
```
paul-graham, line 22: Is something going to go in here at some point in the future? If not, it's not needed. The same for the constructor.

connorpgpmcelroy (approved):

paul-graham (commented):
Little bit of duplication in `info-message-component.html` due to the switch statement. Might be better to have the banner class, SVG path, and assistive message come out of the TypeScript. Everything else looks to be identical...?

src/work-allocation/components/info-message/info-message.component.ts
ficklephil: Good spot, left in accidentally.

paul-graham (approved):
All looks good to me. :)

PR #779, opened 2020-12-07
EUI-2766: Task Manager

paul-graham:
### JIRA link ###
https://tools.hmcts.net/jira/browse/EUI-2766

**Also:**
- https://tools.hmcts.net/jira/browse/EUI-2764
- https://tools.hmcts.net/jira/browse/EUI-2889
- https://tools.hmcts.net/jira/browse/EUI-2961

### Change description ###
The Task Manager component with filtering by location and caseworker.

```
[ ] Yes
[x] No
```


connorpgpmcelroy (approved):

src/work-allocation/services/session-storage.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from '@angular/core';
```
udaydenduluri33, line 1: Should this be moved to common-library at some point?

udaydenduluri33 (approved):
LGTMp

src/work-allocation/services/session-storage.service.ts
paul-graham: If it's going to be used elsewhere in MC, it should be moved to a shared area within MC. If it's going to be used in one of the other applications, the common library makes sense.

For now, though, this is the only use of session storage I've found, which is why it's sitting where it is. :)

udaydenduluri33 (commented):
Flawless

PR #780, opened 2020-12-07
allow static caching of assets

paddybasi:

### Change description ###

```
[ ] Yes
[ ] No
```


Daniel-Lam (approved):
Approved, @paddybasi

andywilkinshmcts (changes requested):
paddy - I think I might only be complaining about the title of the PR, but surely this PR removes cache control headers for static assets, meaning that the effect is to "add caching for static assets" , which kind of is the opposite to the title.

Also how does this interact with the IIS config in web.config which is adding these headers?

paddybasi:
> paddy - I think I might only be complaining about the title of the PR, but surely this PR removes cache control headers for static assets, meaning that the effect is to "add caching for static assets" , which kind of is the opposite to the title.
> 
> Also how does this interact with the IIS config in web.config which is adding these headers?

@andywilkinshmcts yes you're right, changed the title to reflect. The web.config file is actually no longer used as far as I'm aware, that was only necessary when we were using azure webapps; now we've moved to AKS we can/should remove this? @ernestman28 can you confirm please?

andywilkinshmcts (approved):

Mo-Lala84 (approved):
approved

Mo-Lala84 (approved):
lgtm

PR #782, opened 2020-12-08
Update ccd-case-ui-toolkit for ET perf issue

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-2741

### Change description ###
Improve performance time when loading a case with large hearing and judgement collections

```
[ ] Yes
[x] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@angular/platform-browser": "^7.2.13",
     "@angular/platform-browser-dynamic": "^7.2.13",
     "@angular/router": "^7.2.13",
-    "@hmcts/ccd-case-ui-toolkit": "^2.65.6",
+    "@hmcts/ccd-case-ui-toolkit": "^2.66.0-performance-issue-tribunals",
     "@hmcts/ccpay-web-component": "3.0.3",
     "@hmcts/frontend": "0.0.39-alpha",
     "@hmcts/media-viewer": "2.5.1",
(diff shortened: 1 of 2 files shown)


udaydenduluri33 (approved):
LGTMp

package.json
```diff
@@ -191,15 +191,11 @@
...
     "tslint": "~5.11.0",
     "typescript": "~3.2.2",
+    "uuid": "^8.3.1",
     "webpack": "^4.41.5",
     "webpack-cli": "^3.3.10",
     "webpack-node-externals": "^1.7.2",
-    "webpack-source-map-support": "^2.0.1",
-    "base-64": "^0.1.0",
-    "node-fetch": "^2.6.1",
-    "mocha": "^8.1.1",
```
adnanakgun, line 200: Can you make sure that node-fetch and mocha are still in the package.json please?

adnanakgun (approved):

package.json
paulhowes-HMCTS: node-fetch and mocha appear as devDependencies. isomorphic-fetch has a dependency on node-fetch so I assume that is why node-fetch has been removed from the dependencies? As for mocha - it's a testing tool so is that why it has automatically been removed from dependencies? Following on from our conversation we agreed we should check it does not break anything prior to QA.

PR #795, opened 2020-12-15
EUI-2897 Refresh Task and 400 Messages shown at the same time as per design. Messages removed when user clicks on diff sub nav item.

ficklephil:
Two fixes/features.

1. We now use an array of messages, so that messages can be stacked on top of each other in the view, as they are added.

Why? - This comes into play for a business requirement, that requires both a 400 message and a 'Your task list has been refreshed' message to be shown one after the other, on top of each other as per the UI designs.

Use case: 
The 400 error message is shown first, we then automatically refresh the Available Tasks list, if the ATL has been refreshed successfully, we then show the 'Your task list has been refreshed' message under the 400 error message.

2. Messages removed when the user clicks on a different sub-navigation item.


paul-graham (approved):
Looks good to me. :)

paul-graham (approved):
Re-approving after the change to remove messages when navigating.

src/work-allocation/containers/task-list-wrapper/task-list-wrapper.component.ts
```diff
@@ -98,6 +99,31 @@ export class TaskListWrapperComponent implements OnInit {
...
+      const message: InformationMessage = {
+        type: InfoMessageType.INFO,
+        message: InfoMessage.LIST_OF_AVAILABLE_TASKS_REFRESHED,
+      };
+
+      this.infoMessageCommService.addMessage(message);
+    });
+  }
+
+  public getTasks(): Observable {
```
paul-graham, line 121: I think this might be better named as something like "performSearch()` as the current name is easy to confuse with `loadTasks()`.

udaydenduluri33 (changes requested):
@ficklephil  looks like this is service which is using state. Not sure this is the right way to have a state inside a service. @ronaldmansveld  had a different idea for implementing this. Could you please get in touch with @ronaldmansveld  and run it through him before we can approve this? Thanks.

udaydenduluri33:
Also apologies for coming in late on this as you know I was looking at performance related issues last week.

ronaldmansveld (approved):

udaydenduluri33 (approved):
LGTMp

src/work-allocation/containers/task-list-wrapper/task-list-wrapper.component.ts
ficklephil: Agreed.:)

PR #801, opened 2020-12-17
EUI-2913: Reassign task manager

paul-graham:
### JIRA links ###
https://tools.hmcts.net/jira/browse/EUI-2769
https://tools.hmcts.net/jira/browse/EUI-2770
https://tools.hmcts.net/jira/browse/EUI-2795
https://tools.hmcts.net/jira/browse/EUI-2838
https://tools.hmcts.net/jira/browse/EUI-2840
https://tools.hmcts.net/jira/browse/EUI-2913
https://tools.hmcts.net/jira/browse/EUI-2953

### Change description ###
The reassignment page now returns from whence it came when cancelling, succeeding, or with certain errors, rather than always going to My tasks.

Changed the mechanism for setting up actions on returned tasks so it's now done dynamically in the node layer, rather than being returned from the API itself.

This change to the mechanism also necessitated some changes in the way we make the calls so several Angular components and containers had to have their calls to the API modified.

```
[ ] Yes
[x] No
```


connorpgpmcelroy (approved):
There was a lot to investigate but it all looks good, at least from my understanding of it

udaydenduluri33 (approved):
LGTMp

PR #821, opened 2021-01-12
Fix caselist loading when not running on ES

banderous:
### Change description ###

This fixes loading of the case list when not running on Elastic Search.

This change avoids returning jurisdicitonal details instead of case data.

```
[ ] Yes
[x] No
```


Mo-Lala84 (approved):
approved

ernestman28 (approved):

PR #835, opened 2021-01-19
Ensured that No cases message appears AFTER a search and not whilst page is loading.

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-2029

### Change description ###
Ensured that the 'No cases…' messages does not show unless a search has been performed.
Fixed an issue whilst in the area with the IsCaseShareVisible$ observable which was causing a changeDetect error

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/cases/containers/case-list/case-list.component.ts
+++ b/src/cases/containers/case-list/case-list.component.ts
@@ -14,7 +14,7 @@ import { DefinitionsService } from '@hmcts/ccd-case-ui-toolkit/dist/shared/servi
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
 import { SharedCase } from '@hmcts/rpx-xui-common-lib/lib/models/case-share.model';
 import { select, Store } from '@ngrx/store';
-import { BehaviorSubject, combineLatest, Observable, Subscription } from 'rxjs';
+import { BehaviorSubject, combineLatest, forkJoin, Observable, Subscription } from 'rxjs';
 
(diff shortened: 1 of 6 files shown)


src/cases/containers/case-list/case-list.component.ts
```diff
@@ -234,6 +237,8 @@ export class CaseListComponent implements OnInit, OnDestroy {
       this.onPaginationSubscribeHandler(paginationDataFromResult);
     }
 
+    if (typeof resultView.results !== 'undefined') this.resultViewIsReady = true;
```
paul-graham, line 240: This will only ever be evaluated when the containing method is first called. Is that the desired behaviour or should it go back to false if `resultView.results` is `undefined` at a later point?

paul-graham: Same for the one below...

phillip-whitaker: This is desired. The issue is that the resultView was being used in the template as an if, defaulted to null. But with the ngrx store, when the component first loads it gets an empty object  back. This being a truth value displayed the ccd-search result before the object had been properly hydrated. The effect of this was that the search-result component believed that a search had occurred and there were no results, so shows the "No cases found..." message. 

So what I have done here is to make sure that the search-result does not show until the object has been hydrated into a proper SearchResultView object

paul-graham (approved):

PR #902, opened 2021-02-18
Rename the button text

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EUI-2034

### Change description ###
Rename button text and replace button with anchor tag

```
[ ] Yes
[x] No
```


src/cases/containers/case-filter/case-filter.component.ts
```diff
@@ -36,7 +36,7 @@ export class CaseFilterComponent implements OnInit {
 
   ngOnInit(): void {
     this.fromCasesFeature = fromCasesFeature;
-    this.startButtonText = 'Start'; // TODO add this to some config file.
+    this.startButtonText = 'Start now'; // TODO add this to some config file.
```
paul-graham, line 39: Is this still not going into a config file then?

RiteshHMCTS: As discussed only text change is mentioned in the ticket

adnanakgun: Can you please move this string into the app.constants file @RiteshHMCTS ?

RiteshHMCTS: I have moved the text to app.constants file

src/app/app.constants.ts
```diff
@@ -205,6 +205,8 @@ const appHeaderTitle = {name: 'Manage cases', url: '/'};
 // Making the base URL for case details a constant.
 const caseDetailsUrl: string = '/cases/case-details/';
 
+// Create Case button text
+const buttonText: string = 'Start now';
```
adnanakgun, line 209: Sorry to be picky, but it might be better to have a more descriptive name for this. Such as `startButtonText`... `buttonText` is quite vague.

adnanakgun (changes requested):

src/app/app.constants.ts
RiteshHMCTS: Done

adnanakgun (dismissed):

paul-graham (dismissed):

PR #909, opened 2021-02-22
Feature/e UI 3225 wildcard search

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-3396

### Change description ###
Wildcard search

- Wildcard searches based on role type. 
- Wild card search based on case type and case type fields

```
[ ] Yes
[x ] No
```

### DIFF
--- a/api/searchCases/index.ts
+++ b/api/searchCases/index.ts
@@ -5,6 +5,8 @@ import {WILDCARD_SEARCH_FIELDS, WILDCARD_SEARCH_ROLES} from '../configuration/re
 import {fieldNameMapper} from '../lib/util'
 import {ElasticSearchQuery} from './interfaces/ElasticSearchQuery'
 
+const specialCharacters: string[] = [' ', '-', '_']
+
 /**
  * Manually creating Elastic search query
  */
@@ -192,6 +194,5 @@ function canApplyWildCardSearch(wildcardSearchFields: { [key: string]: string[] 
 }
 
 function phraseHasSpecialCharacters(phrase: string): boolean {
-  const specialCharacters: string[] = [' ', '-', '_']
   return specialCharacters.filter((specialCharacter: string) => phrase.indexOf(specialCharacter) >= 0).length > 0
 }
--- a/api/searchCases/index.ts
+++ b/api/searchCases/index.ts
@@ -5,7 +5,7 @@ import {WILDCARD_SEARCH_FIELDS, WILDCARD_SEARCH_ROLES} from '../configuration/re
 import {fieldNameMapper} from '../lib/util'
 import {ElasticSearchQuery} from './interfaces/ElasticSearchQuery'
 
-const specialCharacters: string[] = [' ', '-', '_']
+const SPECIAL_CHARACTERS: string[] = [' ', '-', '_']
 
 /**
  * Manually creating Elastic search query
@@ -194,5 +194,5 @@ function canApplyWildCardSearch(wildcardSearchFields: { [key: string]: string[] 
 }
 
 function phraseHasSpecialCharacters(phrase: string): boolean {
-  return specialCharacters.filter((specialCharacter: string) => phrase.indexOf(specialCharacter) >= 0).length > 0
+  return SPECIAL_CHARACTERS.filter((specialCharacter: string) => phrase.indexOf(specialCharacter) >= 0).length > 0
 }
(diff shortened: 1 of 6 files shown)


api/searchCases/index.ts
```diff
@@ -1,137 +1,166 @@
...
-            caseCriteria = {
-                ...caseCriteria,
-                [newKey]: metaCriteria[key],
-            }
-            delete metaCriteria[key]
-        }
-    })
+export function userCanPerformWildCardSearch(userInfo: UserInfo): boolean {
+  const allowedRoles = getConfigValue(WILDCARD_SEARCH_ROLES) as string[]
+  return userInfo.roles.filter((role: string) => allowedRoles
```
paul-graham, line 32: Might be sensible to do a check to make sure you actually have a `userInfo` object that actually has `roles` at this point - I've seen instances where the user hasn't been set by the time someone tries to do something with it in this codebase already.

api/searchCases/index.ts
```diff
@@ -1,137 +1,166 @@
...
+    .map((allowedRole: string) => allowedRole.toLowerCase())
+    .indexOf(role.toLowerCase()) >= 0).length > 0
```
paul-graham, line 34: Is `indexOf(...) >= 0` the desired approach? That means that `'caseworker'` can appear _anywhere_ in the string. Is that what they're after or should it be a prefix? Just a question, not saying it's wrong...

paul-graham (changes requested):
Reviewed what I can but `api/searchCases/index.ts` has changed enormously due to a reformat so it's difficult to pick out what has changed.

api/searchCases/index.ts
kuda1992: @paul-graham are we saying we have some instances where the user roles are defined?

api/searchCases/index.ts
kuda1992: Yes thats exactly what I want. The user should just have the role required to perform a wildcard search

Mo-Lala84 (approved):
approved

api/searchCases/index.ts
paul-graham: It's possible. We've seen race conditions with user profiles in the past. I don't know if they're all resolved or quite how this fits into the sequence. Either way, it's always sensible to do null checking on parameters.

api/searchCases/index.ts
```diff
@@ -1,137 +1,171 @@
...
+  }
+  return sortQuery
 }
 
 function isKeywordSuffixNeeded(columnName, type): string {
-    const types = ['Text', 'TextArea', 'FixedList', 'FixedListEdit', 'MultiSelectList', 'FixedRadioList', 'DynamicList']
-    const specialFields = ['reference', 'jurisdiction', 'state', 'case_type_id']
-    const isText = types.includes(type) || specialFields.includes(columnName)
-    return isText ? '.keyword' : ''
+  const types = ['Text', 'TextArea', 'FixedList', 'FixedListEdit', 'MultiSelectList', 'FixedRadioList', 'DynamicList']
```
paul-graham, line 165: These should be defined outside of this instance method as they'll be instantiated every time the method is invoked. Not a blocker - and I know you've just changed the indenting - but it's less efficient than it could be.

api/searchCases/index.ts
```diff
@@ -144,14 +178,20 @@ export function handleElasticSearchResponse(proxyRes, req, res, json): {} {
...
   }
+  return {}
+}
+
+function canApplyWildCardSearch(wildcardSearchFields: { [key: string]: string[] }, caseType: string, criterion: string): boolean {
+  return wildcardSearchFields.hasOwnProperty(caseType) && wildcardSearchFields[caseType].indexOf(criterion) >= 0
+}
+
+function phraseHasSpecialCharacters(phrase: string): boolean {
+  const specialCharacters: string[] = [' ', '-', '_']
```
paul-graham, line 195: This should also be declared outside of this instance method.

paul-graham (approved):
A couple of things that would benefit from a bit of a clean-up but otherwise looks fine to me - they're not blockers.

One other thing to mention - a lack of semicolons is now a lint warning. We shouldn't be introducing even more code without semicolons. For future changes in the node layer, please ensure you add semicolons appropriately or the list of lint warnings will simply increase.

api/searchCases/index.ts
kuda1992: Sure next time I look at it, I will fix it and I didn't modify this code.

udaydenduluri33 (approved):
LGTMk

api/searchCases/index.ts
kuda1992: @paul-graham moved the special character array outside the scope of the function.

PR #913, opened 2021-02-23
Feature/e UI 3382 assign changes

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-3382

### Change description ###
In Task Assign page, pre-selecting location in location dropdown via user's location. Also, changing the caseworkers selectable based on the location selected

```
[ ] Yes
[ x ] No
```


api/workAllocation/interfaces/task.ts
```diff
@@ -60,10 +60,25 @@ export interface Location {
...
 }
 
+export interface LocationApi {
+    location_id: string
+    location: string
+    is_primary: boolean
+    services: string[]
+}
+
 export interface Caseworker {
```
paul-graham, line 70: What is this one for now that there's a `CaseworkerApi` interface that looks like it replaces it?

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -1,5 +1,8 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { Router } from '@angular/router';
+import c = require('config');
+import { UserInfo } from 'src/app/models/user-details.model';
```
paul-graham, line 4: This import and the one below shouldn't begin with `src/`.

src/work-allocation/components/task-assignment/task-assignment.component.spec.ts
```diff
@@ -2,6 +2,7 @@ import { Component, Input, ViewChild } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { Router } from '@angular/router';
 import { Observable } from 'rxjs';
+import { SessionStorageService } from 'src/app/services';
```
paul-graham, line 5: This import shouldn't begin with `src/`.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -103,6 +116,34 @@ export class TaskAssignmentComponent implements OnInit {
...
+  /**
+   * Sets the logged in user id
+   */
+  public setUserId(): void {
+    const userInfoStr = this.sessionStorageService.getItem('userDetails');
+    if (userInfoStr) {
+      const userInfo: UserInfo = JSON.parse(userInfoStr);
+      this.userId = userInfo.id;
+    }
+    return undefined;
```
paul-graham, line 128: If your return type is `void`, you shouldn't be returning anything. Is this statement needed or do you need to change your return type?

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -103,6 +116,34 @@ export class TaskAssignmentComponent implements OnInit {
...
+    this.caseworkerService.getAll().subscribe(caseworkers => {
+      const assignedCaseworker = caseworkers.find(cw => this.isAssignedCaseworker(cw.idamId));
+      this.assignedCaseworker = assignedCaseworker;
+    }, error => {
+      handleFatalErrors(error.status, this.router, WILDCARD_SERVICE_DOWN);
+    });
+  }
+
+  public isAssignedCaseworker(idamId: string): boolean {
+    return idamId === this.userId ? true : false
```
paul-graham: No need for this to be a ternary operator: `return idam === this.userId;` is sufficient.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -83,17 +86,27 @@ export class TaskAssignmentComponent implements OnInit {
   private pExcludeCaseworkers: Caseworker[];
   private readonly caseworkerDisplayName: CaseworkerDisplayName = new CaseworkerDisplayName();
 
+  public userId: string;
```
paul-graham, line 89: Why are these `public` properties? Is it purely for unit testing as I don't see them being used anywhere other than the spec file?

If they're _only_ public for unit test purposes, I think some refactoring is needed. The same applies to some of the public methods below. Methods and properties shouldn't be public purely for the benefit of unit testing as it means you're exposing internal workings - and things that are being _used_ as private properties - when you didn't necessarily intend to.

You could, for example, add a getter method for a `private pUserId: string;` property if you need to access it in the unit tests. That, at the least, won't make the private property editable externally. Also, you have a `setUserId()` method, which making this public essentially circumvents. That method is _also_ `public`, which looks less concerning but still appears to be purely for unit testing purposes.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -135,18 +175,24 @@ export class TaskAssignmentComponent implements OnInit {
...
+        const locationCaseWorkers = caseworkers.filter(cw => this.locationMatches(cw));
+        this.setupCaseworkers(locationCaseWorkers);
       }, error => {
         handleFatalErrors(error.status, this.router, WILDCARD_SERVICE_DOWN);
       });
     }
   }
 
+  // check if the caseworker's location exists and matches a location
+  public locationMatches(cw: Caseworker): boolean {
```
paul-graham, line 188: The name of this method doesn't _really_ describe what it's for - if you're checking that a caseworker's location matches the currently selected location, `locationMatches` doesn't really tell the story.

Also, `public` methods (and I'm also not convinced this should be `public` - see other comment as to why) should have `/** */` style comments so that they're properly documented when attempting to use them from another component.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -159,4 +205,11 @@ export class TaskAssignmentComponent implements OnInit {
     return toVet;
   }
 
+  // sets the selected location on the location dropdown
+  public setSelectedLocation(): void {
```
paul-graham, line 209: Same comments as the `locationMatches` method - it _looks_ like you're actually choosing the location of the assigned caseworker, not just generically setting a selected location. My initial reading of this method was that it was getting called from the HTML or something but then I realised it also didn't have a parameter.

On that subject - `setX` methods should take a parameter with the value you're trying to set a property to. I think you're looking for `setupX` in this instance.

paul-graham (changes requested):

api/workAllocation/interfaces/task.ts
connorpgpmcelroy: So the CaseworkerApi is the type that refers to the caseworker object given via the API, Caseworker is the class that gets the properties via the CaseworkerApi in order to return a Caseworker class object. So non-specified API information -> CaseworkerApi -> Caseworker (same with LocationApi and Location).

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Ok, will change

src/work-allocation/components/task-assignment/task-assignment.component.spec.ts
connorpgpmcelroy: Will change

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Will remove this return statement

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Yes, I noticed that and changed it already ten mins ago when I was looking through code. It was probably both a remnant of checking that the userId was present and a lapse in concentration

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Yes, I agree with that. I'll rework the public properties into private ones (similar to the other properties within the file) and refactor the tests.

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Ok, /** */ type comment will be added (and method name changed). Will also make private

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Yep, will change to setupSelectedLocation()

api/workAllocation/interfaces/task.ts
paul-graham: Is there any way we can just have one interface? Is it simply because of *camel vs snake*? If so, we could make the appropriate changes wherever the interfaces are used within Angular and do away with the need for translation from one to another.

That could go in a separate tech debt ticket but definitely worth looking at, I think.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -75,25 +78,67 @@ export class TaskAssignmentComponent implements OnInit {
...
+   * Gets the private logged in user id
+   */
+  public get userId(): string {
+    return this.pUserId;
+  }
+
+  /**
+   * Sets the private logged in user id
+   */
+  public set userId(value: string) {
```
paul-graham, line 91: I still think this circumvents the `setupUserId()` method, which isn't ideal behaviour. It provides a way for the ID of user stored in the session and the `userId` in here to get out of sync, which could easily lead to problems. My prior comment was to highlight this risk, rather than to suggest using getters and setters instead. 🤷

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -75,25 +78,67 @@ export class TaskAssignmentComponent implements OnInit {
...
+   * Gets the private logged in user id
+   */
+  public get assignedCaseworker(): Caseworker {
+    return this.pAssignedCaseworker;
+  }
+
+  /**
+   * Sets the private logged in user id
+   */
+  public set assignedCaseworker(value: Caseworker) {
```
paul-graham, line 107: Same applies to this - this will allow the assigned caseworker to be overridden and circumvent `setupAssignedCaseworker()`. In this instance, it looks like that might be not be a terrible idea - i.e., you can assign to someone else - but, again, this _looks_ like it's just for the benefit of unit tests...?

paul-graham (changes requested):

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Will remove setter to decrease risk

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: Will remove setter (will leave comment stating could be useful)

paul-graham (approved):
Approved. Thank you for making the requested changes, @connorpgpmcelroy.

Note that there remains one unresolved conversation regarding what look like a couple of duplicate (ish) interfaces. Can you raise a JIRA to look at rationalising them to get rid of the need for transformation? :)

api/workAllocation/interfaces/task.ts
connorpgpmcelroy: CaseworkerApi is only used for defining details received via API and is then immediately used to define a Caseworker object (same with LocationApi -> Location). Happy to make a ticket for it to investigate and potentially edit.

src/work-allocation/components/task-assignment/task-assignment.component.ts
```diff
@@ -135,18 +196,30 @@ export class TaskAssignmentComponent implements OnInit {
...
+
+  /**
+   * Sets up the caseworkers, ensuring that excluded ones that are filtered out.
+   * @param caseworkers The initial list of caseworkers
+   */
   private setupCaseworkers(caseworkers: Caseworker[]): void {
-    this.pAllCaseworkers = [ ...caseworkers ];
-    this.pCaseworkers = [ ...caseworkers ].filter(item => {
+    this.pAllCaseworkers = [...caseworkers];
+    this.pCaseworkers = [...caseworkers].filter(item => {
```
kuda1992, line 222: [...caseworkers].filter() is this necessary filter creates a new array anyway.

Mo-Lala84 (approved):
approved

kuda1992 (approved):
LGTM

src/work-allocation/components/task-assignment/task-assignment.component.ts
connorpgpmcelroy: I did not write this code (just reformatted) however could just change [...caseworkers].filter -> pAllCaseworkers.filter if that's suitable Kuda?

kuda1992: If this is possible, yes please.

connorpgpmcelroy: Ok, will do

PR #920, opened 2021-02-24
Feature/e UI 3456 route credential

udaydenduluri33:
Using the route credentials option with the changes to the new node library

https://tools.hmcts.net/jira/browse/EUI-3456

### Change description ###

```
[ ] Yes
[X] No
```


ernestman28 (approved):
LGTM

connorpgpmcelroy (approved):
LGTM

api/auth/index.ts
```diff
@@ -56,6 +58,15 @@ export const getXuiNodeMiddleware = () => {
     const idamApiPath = getConfigValue(SERVICES_IDAM_API_URL)
     const s2sSecret = getConfigValue(S2S_SECRET)
     const tokenUrl = `${getConfigValue(SERVICES_IDAM_API_URL)}/oauth2/token`
+    const userName = getConfigValue(SYSTEM_USER_NAME)
+    const password = getConfigValue(SYSTEM_USER_PASSWORD)
+
+    const routeCredential = {
```
paul-graham, line 64: Can we make this property name a bit more descriptive if it's specific to the caseworker route? At the moment, it's quite generic and I _assume_ it's not going to be used for any other routes...?

Jenkinsfile_CNP
```diff
@@ -31,7 +31,9 @@ def secrets = [
     secret('webapp-redis-connection-string', 'REDISCLOUD_URL'),
     secret('test-email', 'TEST_EMAIL'),
     secret('test-password', 'TEST_PASSWORD'),
-    secret('appinsights-instrumentationkey-mc', 'APPINSIGHTS_INSTRUMENTATIONKEY')
+    secret('appinsights-instrumentationkey-mc', 'APPINSIGHTS_INSTRUMENTATIONKEY'),
+    secret('system-user-name', 'SYSTEM_USER_NAME'),
```
paul-graham, line 35: Which system? Are these generic credentials that will work across the board (which seems a bit risky) or are they for a particular service? If the latter, a more descriptive name would probably be better. The same applies to the password, below.

charts/xui-webapp/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: xui-webapp
 home: https://github.com/hmcts/rpx-xui-webapp
-version: 0.2.9
+version: 0.2.10
```
paul-graham, line 4: Intentional? This a recurring theme so I'm guessing something is happening during a `yarn install`. Can we clarify which version we should be on? If we need to stick to `0.2.9`, we need to fix it at that in the `package.json`.

api/configuration/references.ts
```diff
@@ -60,6 +60,10 @@ export const JURISDICTIONS = 'jurisdictions'
 
 export const IDAM_SECRET = 'secrets.rpx.mc-idam-client-secret'
 
+export const SYSTEM_USER_NAME = 'secrets.rpx.system-user-name'
```
paul-graham, line 63: Same comment as above about this being a generic name...

yarn.lock
```diff
@@ -485,10 +485,10 @@
   dependencies:
     tslib "^1.9.0"
 
-"@hmcts/rpx-xui-node-lib@^2.24.3":
-  version "2.24.3"
-  resolved "https://registry.yarnpkg.com/@hmcts/rpx-xui-node-lib/-/rpx-xui-node-lib-2.24.3.tgz#63292f848fb10a475d7aa65abeefa0b030a5ed9f"
-  integrity sha512-5pmLQIdJ3CdpxzMa+6IGFQ4+MB24yZ9Ocim7DlLvTe2jHQYZwE/TlbNGkhRjXmqPQy4bt/9zuk4X8IhyDgxtdA==
+"@hmcts/rpx-xui-node-lib@^2.24.5":
+  version "2.25.0"
```
paul-graham, line 489: Is this the version we want? You seem to have specifically changed it in `package.json` to `^2.24.5` from `^2.24.3` but it's actually installed `2.25.0`. This looks like an unintentional change.

paul-graham (changes requested):

yarn.lock
udaydenduluri33: This is intentional change

charts/xui-webapp/Chart.yaml
udaydenduluri33: Jenkins automatically does this. I did not do. Its needed when it notices a change in Values.yaml

api/configuration/references.ts
udaydenduluri33: we want it to be generic. This is a system user name which we are going to use it for several roles/services.

api/auth/index.ts
udaydenduluri33: we want it to be generic. This is a system user name which we are going to use it for several roles/services.

Jenkinsfile_CNP
udaydenduluri33: we want it to be generic. This is a system user name which we are going to use it for several roles/services.

api/auth/index.ts
paul-graham: But this only seems to apply to a single route. Will we be adding more routes in the future with identical credentials?

yarn.lock
paul-graham: So why specify `2.24.5` in the package.json? If you _want_ `2.25.0`, why not specify that?

api/auth/index.ts
udaydenduluri33: Yes you got the point. This is open to other routes.

yarn.lock
udaydenduluri33: I get your question now. I tried it explicitly changing in yarn.lock. But, it moves it back to `"@hmcts/rpx-xui-node-lib@^2.24.5"` Its autogenerated by running yarn.

paul-graham: Yeah, you can't change it in `yarn.lock`. It's happening when you do `yarn [install]` because it's set to `"^2.24.5"` in the `package.json`. The caret prefix ("^") indicates that it should use the latest compatible version so it'll get any newer minor or patch releases.

That _may_ ultimately be what we want to do but I'm guessing you chose `2.24.5` because you made changes in that version that you want to be part of this release. And you _didn't_ make changes in any versions from `2.24.6` onwards (or, if you did, they'll be in a different PR)...?

The safest way to test those changes in isolation is to remove the caret prefix in the `package.json` file and that's what I'd recommend.

paul-graham: Thoughts on this, @Mo-Lala84?

udaydenduluri33: This is resolved now

paul-graham (approved):
All concerns addressed and necessary changes made. 👍

PR #926, opened 2021-02-25
added logic for deep linking

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-1171 

### Change description ###

Created a redirect strategy for deep linking.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/app/services/auth/auth.guard.spec.ts
+++ b/src/app/services/auth/auth.guard.spec.ts
@@ -5,6 +5,7 @@ import {HttpClient} from '@angular/common/http';
 import { StoreModule } from '@ngrx/store';
 import { of } from 'rxjs';
 import {AuthGuard} from './auth.guard';
+import { SessionStorageService } from '../session-storage/session-storage.service';
 
 
 class HttpClientMock {
@@ -21,6 +22,7 @@ describe('AuthGuard', () => {
       ],
       providers: [
         AuthService,
+        SessionStorageService,
         { provide: HttpClient, useClass: HttpClientMock },
       ]
     });
--- a/src/app/services/session-storage/session-storage.service.spec.ts
+++ b/src/app/services/session-storage/session-storage.service.spec.ts
@@ -20,6 +20,13 @@ describe('Session storage service', () => {
     expect(sessionStorage.setItem).toHaveBeenCalledWith('exampleKey', 'exampleValue');
   });
 
+  it('should remove an item after reading it when remove is set to true', () => {
+    const service = new SessionStorageService();
+    service.getItem('exampleKey');
+    expect(sessionStorage.getItem).toHaveBeenCalledWith('exampleKey');
+    expect(sessionStorage.removeItem).toHaveBeenCalledWith('exampleKey');
+  });
+
   it('should allow getting the item', () => {
     const service = new SessionStorageService();
     service.getItem('exampleKey');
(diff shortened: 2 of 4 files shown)


src/app/services/session-storage/session-storage.service.ts
```diff
@@ -7,9 +7,14 @@ export class SessionStorageService {
 
   /**
    * Get an item from the session storage.
+   * If remove is true, the item will be removed once read
    */
-  public getItem(key: string): string {
-    return sessionStorage.getItem(key);
+  public getItem(key: string, remove: boolean = false): string {
```
phillip-whitaker, line 12: I know it might be questioned why I have added the option.

I want the value to removed after it has been read. I could do this by calling getItem() and then removeItem() in my calling code. But that makes me nervous. I want the read and delete to be done in one, atomic request. I could have added a new method getAndDeleteItem() but that seemed an odd thing to do.

src/app/services/auth/auth.guard.spec.ts
```diff
@@ -5,6 +5,7 @@ import {HttpClient} from '@angular/common/http';
 import { StoreModule } from '@ngrx/store';
 import { of } from 'rxjs';
 import {AuthGuard} from './auth.guard';
```
paul-graham, line 7: Can I be awkward and ask you to clean up the imports up here? This one looks out of place now. :)

src/app/services/session-storage/session-storage.service.ts
paul-graham: Logically, no problem with this. My only slight concern is with the name of the property as I don't think it's _entirely_ clear what the behaviour will be. Renaming the `remove` parameter to `removeAfterRead` would make it much clearer.

src/app/services/session-storage/session-storage.service.ts
```diff
@@ -7,9 +7,14 @@ export class SessionStorageService {
 
   /**
    * Get an item from the session storage.
+   * If remove is true, the item will be removed once read
```
paul-graham, line 10: Maybe prefix this line with `@param removeAfterRead`?

paul-graham (commented):
I've added comments. None of them are blockers but my suggestions are for code readability.

src/app/services/session-storage/session-storage.service.spec.ts
```diff
@@ -20,6 +20,13 @@ describe('Session storage service', () => {
     expect(sessionStorage.setItem).toHaveBeenCalledWith('exampleKey', 'exampleValue');
   });
 
+  it('should remove an item after reading it when remove is set to true', () => {
+    const service = new SessionStorageService();
```
adnanakgun, line 24: Should we not mock this service? You are adding a side effect to the test. In fact, would it not make more sense to test the functions `storeRedirectUrl` and `redirectToStoredUrl` and see if the mock service is called from within those?

phillip-whitaker: With this, I did just copy and paste the test above and added the clause. But I agree with your wider statement that more tests should be added.

adnanakgun: Oops, my bad, I thought you are testing `AuthGuard`. Maybe test `AuthGuard` as well then.

adnanakgun: Sorry

paul-graham (approved):
Looks like a nice, neat solution, sir. :)

adnanakgun (approved):

PR #930, opened 2021-03-02
Feature/e UI 3428 assign and go to case

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-3428

### Change description ###
Fixing functionality to 'assign and go to case'. Also, as a side effect have fixed caseReference links (shown with blue underline)

```
[ ] Yes
[ x ] No
```


src/work-allocation/containers/available-tasks/available-tasks.component.ts
```diff
@@ -79,10 +79,9 @@ export class AvailableTasksComponent extends TaskListWrapperComponent {
    */
   public claimTaskAndGo(task: Task): void {
     this.taskService.claimTask(task.id).subscribe(() => {
-      // constant below removes spaces from caseReference to get caseId
-      const caseId = task.caseReference.replace(/\s/g, '');
+      const goToCaseUrl = `/cases/case-details/${task.caseReference}`
```
paul-graham, line 82: This looks risky. I thought the case reference was a _formatted_ version of the case ID...?

paul-graham (changes requested):
I thought case reference was a formatted version of the case ID...?

src/work-allocation/containers/available-tasks/available-tasks.component.ts
connorpgpmcelroy: Currently the data being passed through isn't formatted at all - we get a 16 digit number with no spaces from API which is currently being used. If we want to format it I can add some functionality to add spaces (similar to the formatting we had before with the data 'XXXX XXXX XXXX XXXX') and use the replace again. We don't actually get a 'caseReference' from the API, we only get a 'case_id' which has no formatting.

paul-graham: Hmmm... then I think we should continue to refer to it as a `caseId` (or `case_id`) rather than `caseReference` as that's what it is. 🤷‍♂️

paul-graham (changes requested):

kuda1992 (approved):
LGTM

udaydenduluri33 (approved):
LGTMc

paul-graham (approved):
Looks good. And well done for lining up all the comments in `task-field-type.ts`! :)

PR #946, opened 2021-03-10
EUI-2667: Collection date fields not displayed

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2667 ✅ `DONE`

### Change description ###
Using a fully released version of the CCD Case UI Toolkit, having passed QA on a tagged version.

```
[ ] Yes
[x] No
```

**Notes**
The ticket was previously marked as `DONE` but the toolkit PR went stale after it was never merged. Things have moved on a fair bit since then so this is to retest that the bug is fixed in a version that was rebased from `master`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@angular/platform-browser": "^7.2.13",
     "@angular/platform-browser-dynamic": "^7.2.13",
     "@angular/router": "^7.2.13",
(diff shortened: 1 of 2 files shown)


adnanakgun (approved):

paul-graham:
A newer version of the CCD Case UI Toolkit has already been deployed, making this PR redundant.

PR #953, opened 2021-03-12
Bug fix for Issue with caseworkers failing when logged in as case officer

udaydenduluri33:

https://tools.hmcts.net/jira/browse/EUI-3602

### Change description ###

Issue with caseworkers failing when logged in as case officer

```
[ ] Yes
[X] No
```

### DIFF
--- a/api/workAllocation/index.ts
+++ b/api/workAllocation/index.ts
@@ -63,8 +63,7 @@ export async function searchTask(req: EnhancedRequest, res: Response, next: Next
     res.status(status);
     // Assign actions to the tasks on the data from the API.
     if (data) {
-      const caseworkers: Caseworker[] = await retrieveAllCaseWorkers(req, res);
-      assignActionsToTasks(data.tasks, req.body.view, caseworkers);
+      assignActionsToTasks(data.tasks, req.body.view, req.session.caseworkers);
     }
 
     // Send the (possibly modified) data back in the Response.
--- a/api/workAllocation/util.ts
+++ b/api/workAllocation/util.ts
@@ -75,7 +75,9 @@ export function assignActionsToTasks(tasks: any[], view: any, caseworkers: Casew
       task.taskName = task.name
       task.caseName = task.case_name
       task.caseCategory = task.case_category
-      task.assigneeName = getAssigneeName(task, caseworkers);
+      if (caseworkers) {
+        task.assigneeName = getAssigneeName(task, caseworkers);
+      }
     }
   }
 }


api/workAllocation/index.ts
```diff
@@ -63,8 +63,7 @@ export async function searchTask(req: EnhancedRequest, res: Response, next: Next
     res.status(status);
     // Assign actions to the tasks on the data from the API.
     if (data) {
-      const caseworkers: Caseworker[] = await retrieveAllCaseWorkers(req, res);
-      assignActionsToTasks(data.tasks, req.body.view, caseworkers);
+      assignActionsToTasks(data.tasks, req.body.view, req.session.caseworkers);
```
paul-graham, line 66: I don't understand why this change has been made. The `await retrieveAllCaseWorkers(req, res)` call was already returning it from `req.session.caseworkers` if it was on there and, if not, it was loading the caseworkers.

paul-graham (changes requested):

api/workAllocation/index.ts
udaydenduluri33: This is made because the normal caseworker who is logged in cannot access caseworkers api. Its easier to show rather than type

udaydenduluri33: If I am logged in as case worker the assignee name is not required.
Only if I am logged in as senior case worker the assignee name is required

adnanakgun (approved):

ronaldmansveld (approved):

paul-graham (changes requested):
It has been explained to me by @udaydenduluri33 (on Slack) that the change in `searchTask(...)` is to resolve an issue where the call to retrieve caseworkers will fail for users that don't have permission to retrieve them.

My suggestion remains that the better solution is either:
1. Identify prior to making the call the roles that have permission to retrieve them and only make the call for those roles, or;
2. Handle the first failure gracefully and then make no subsequent calls for the logged in user with the knowledge that they have no permission.

What is _not_ a good idea is to simply bypass making the call and introduce a race condition where you're relying on another call having been made previously. If you have call chaining in place to ensure the first call completes before this call could ever be made, pass the list of already-retrieved caseworkers down to this method rather than relying on an indirect method for access.

Neither manual testing nor automation testing will reliably catch a race condition - the solution _must_ be within the code.

udaydenduluri33:
> * Identify prior to making the call the roles that have permission to retrieve them and only make the call for those roles, or;
> * Handle the first failure gracefully and then make no subsequent calls for the logged in user with the knowledge that they have no permission.

@paul-graham 
Tried to answer that this will not have any race conditional issues. I can explain in detail if you want in a call what this fix is all about.

api/workAllocation/index.ts
paul-graham: Where is that logic in the code?

PR #956, opened 2021-03-15
remove proxy.end() inside of activityTracker

kuda1992:
### Change description ###
Remove proxy.end() inside of activityTracker which is causing post proxy requests to fail

```
[ ] Yes
[ x] No
```


Mo-Lala84 (approved):
approved

PR #972, opened 2021-03-27
EUI-3587: Fix "Print URL" pipe for rewriting remote Print Service URLs

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3587

### Change description ###
Pull request to create a PR environment for testing bugfix for EUI-3587, which fixes the pipe to rewrite remote Print Service URLs to the correct "local" version to be used for the front-end, irrespective of environment-based variations of the remote domain name.

```
[ ] Yes
[x] No
```


src/app/services/ccd-config/ccd-case.config.ts
```diff
@@ -96,10 +96,6 @@ export class AppConfig extends AbstractAppConfig {
     return this.config.print_service_url;
   }
 
-  public getRemotePrintServiceUrl() {
```
paul-graham, line 99: I think this makes it a breaking change...? That would mean it needs to go up a major version.

paul-graham (changes requested):
I think the toolkit needs to have been a major release because it introduces a breaking change in the public API.

src/app/services/ccd-config/ccd-case.config.ts
adnanakgun: Agree with @paul-graham. Would this not break the build?

Daniel-Lam: @paul-graham I've now re-introduced this in `app.config.ts` in the CCD UI Toolkit as a deprecated, no-op function that returns `undefined`, ahead of its removal in the next major release.

adnanakgun (approved):

paul-graham (approved):
All looks good. Definitely best to get ahead of a deprecation and ditch the redundant code. 😃

adnanakgun (approved):

paul-graham (approved):
Happy with this. 👍

PR #977, opened 2021-03-30
Bugfix/e UI 2925 activity tracker proxy 3

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-3638
as well as
https://tools.hmcts.net/jira/browse/EUI-2925

### Change description ###

Toolkit now posts activity every 5 seconds but gets activities every 30 seconds

```
[ ] Yes
[x] No
```


Mo-Lala84 (approved):
approved

udaydenduluri33 (approved):
LGTMa

api/lib/middleware/proxy.ts
```diff
@@ -9,6 +9,13 @@ const logger = log4jui.getLogger('proxy')
 
 export const onProxyError = (err, req, res) => {
     logger.error(err)
+
+    if (req.baseUrl && req.baseUrl === '/activity' && req.user && req.user.userinfo) {
```
paul-graham, line 13: You don't really need the initial `req.baseUrl` test because it's entirely covered by `req.baseUrl === '/activity'`.

src/cases/resolvers/activity.resolver.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should verify if user is authorised for activity on resolve and return true when is authorised', () => {
+    let result: boolean;
+    activityResolver.resolve().subscribe(r => result = r);
+    expect(activityService.verifyUserIsAuthorized).toHaveBeenCalled();
+    expect(result).toBe(true);
+  });
+
+  it('should verify if user is authorised for activity on resolve and return true when is not authorised', () => {
```
paul-graham, line 23: Does this test actually make sense? It should say they're _authorised_ for activity when _not_ authorised? 🤨

src/cases/resolvers/activity.resolver.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class ActivityResolver implements Resolve<boolean> {
+
+  constructor(private readonly activityService: ActivityService) {}
+
+  public resolve(): Observable<boolean> {
+    this.activityService.verifyUserIsAuthorized();
+    return Observable.of(true);
```
paul-graham, line 13: Why does this always return true? Is it not based on whether or not the user is authorised? If not, what is the point of it? 🤷‍♂️

paul-graham (changes requested):
I have questions...

src/cases/resolvers/activity.resolver.spec.ts
adnanakgun: Most of this stuff was copied over from CCD-UI. I'll have look at the test.

src/cases/resolvers/activity.resolver.ts
adnanakgun: I think this another instance of stuff being copied from CCD-UI. `this.activityService.verifyUserIsAuthorized();` where the actual check happens. I think you still would want the resolver to return true regardless of whether the activity call is happening or not.

paul-graham: Why would you still want it to return true even if the user is not authorised? Doesn't that make the call redundant? Or perhaps it's a poorly named method...?

paul-graham: Oh, `this.activityService.verifyUserIsAuthorized()` doesn't return anything. How odd. 🤯

src/cases/resolvers/activity.resolver.spec.ts
paul-graham: Given the code, the test is correct... but seems a bit pointless because `result` cannot ever be anything _other_ than `true`. 😂

src/cases/resolvers/activity.resolver.ts
adnanakgun: If the resolver returns false the navigation would not happen at all, which is not what we want. We only want the activity call to be made if it is authorised to do so. I agree that maybe a resolver is not the best way to do it.

paul-graham (approved):
A resolver might not be the right solution for this but it doesn't seem to cause any harm... 🤷‍♂️

Mo-Lala84 (approved):
approved

udaydenduluri33 (approved):
LGTMa

leepavlouhmcts (dismissed):

test/e2e/support/reports.js
```diff
@@ -1,15 +1,15 @@
-const reporter = require('cucumberjs-allure-reporter');
-const fs = require('fs');
-const mkdirp = require('mkdirp');
+// const reporter = require('cucumberjs-allure-reporter');
```
Daniel-Lam, line 1: Perhaps just remove these commented-out lines, if they're no longer needed?

test/e2e/support/reports.js
```diff
@@ -1,15 +1,15 @@
...
+// const fs = require('fs');
+// const mkdirp = require('mkdirp');
 
-const xmlReports = `${process.cwd()}/reports/xml`;
+// const xmlReports = `${process.cwd()}/reports/xml`;
```
Daniel-Lam, line 5: Same comment as above.

test/e2e/support/reports.js
```diff
@@ -1,15 +1,15 @@
...
 
-// var createXmlDir = (function() {
-//     if (!fs.existsSync(xmlReports)) {
-//         mkdirp.sync(xmlReports);
-//     }
-// })();
+// // var createXmlDir = (function() {
```
Daniel-Lam, line 7: Same for these commented-out lines, too.

test/e2e/support/reports.js
```diff
@@ -1,15 +1,15 @@
...
+// //     if (!fs.existsSync(xmlReports)) {
+// //         mkdirp.sync(xmlReports);
+// //     }
+// // })();
 
-reporter.config({ targetDir: xmlReports });
+// reporter.config({ targetDir: xmlReports });
```
Daniel-Lam, line 13: Same for this commented-out line.

test/e2e/support/reports.js
```diff
@@ -1,15 +1,15 @@
...
 
-module.exports = reporter;
+// module.exports = reporter;
```
Daniel-Lam, line 15: Same for this commented-out line.

test/ngIntegration/tests/stepDefinitions/caselistPage.steps.js
```diff
@@ -86,7 +86,7 @@ defineSupportCode(function ({ And, But, Given, Then, When }) {
 
     Then('I Validate total cases count {string}', async function(workbasketConfigRef){
         const workbasketConfig = global.scenarioData[workbasketConfigRef].getConfig();
-        let reqData = { size: 250 }
+        let reqData = { size: 25 }
```
Daniel-Lam, line 89: Consider using `const` instead of `let`.

test/ngIntegration/tests/stepDefinitions/searchCase.steps.js
```diff
@@ -92,7 +92,7 @@ defineSupportCode(function ({ And, But, Given, Then, When }) {
     Then('I Validate search case total cases count {string}', async function(searchCaseConfigRef){
         const searchCaseConfig = global.scenarioData[searchCaseConfigRef].getConfig();
         
-        let reqData = { size: 250 }
+        let reqData = { size: 25 }
```
Daniel-Lam, line 95: Consider using `const` instead of `let`.

test/nodeMock/ccd/ccdApi.js
```diff
@@ -426,6 +427,21 @@ class CCDApi{
     }
 
 
+    getCaseActivity(cases){
+        let caseIds = cases.split(',');
```
Daniel-Lam, line 431: Consider using `const`s instead of `let`s.

Daniel-Lam (changes requested):
@adnanakgun and @Mo-Lala84 - review complete; just a few minor observations raised.

test/e2e/support/reports.js
Mo-Lala84: yup however we are just validating if the pr actually pass as cucumber was now causing the new errors so temporary

Daniel-Lam: OK, no worries - will mark all these as resolved.

PR #980, opened 2021-03-31
EUI-3664 Fix WA Task Assignee disappearing on refresh

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-3664

### Change description ###
Showing the assignee column had been set via the window.state.history with a boolean value. This fails when a refresh is used as the assignee column disappears. Therefore the assignee column is now set to appear based on which action appears in the url.

```
[ ] Yes
[ x ] No
```


leepavlouhmcts (approved):
looks good

kuda1992 (approved):

udaydenduluri33 (approved):
LGTMc

PR #1009, opened 2021-04-22
Bug/e UI 3621 collection search filter 2 aa

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-3621

### Change description ###

Fixes on MC and new Toolkit

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/cases/services/search-filter.service.ts
+++ b/src/cases/services/search-filter.service.ts
@@ -71,13 +71,16 @@ export class SearchFilterService {
       prefix = parentPrefix + '.';
     }
     for (let attributeName of Object.keys(formGroupValue)) {
-      const value = formGroupValue[attributeName];
+      let value = formGroupValue[attributeName];
       if (isStringOrNumber(value)) {
         const filterType = getFilterType(attributeName, this.metadataFields);
         attributeName = sanitiseMetadataFieldName(filterType, attributeName);
         target[filterType][prefix + attributeName] = value;
       } else if (value) {
-        this.buildFormDetails(attributeName, target, value);
+        if (value[0]) { // is array and has index zero populated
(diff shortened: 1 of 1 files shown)


phillip-whitaker (approved):
LGBTQ+

src/cases/services/search-filter.service.ts
```diff
@@ -71,13 +71,16 @@ export class SearchFilterService {
...
     for (let attributeName of Object.keys(formGroupValue)) {
-      const value = formGroupValue[attributeName];
+      let value = formGroupValue[attributeName];
       if (isStringOrNumber(value)) {
         const filterType = getFilterType(attributeName, this.metadataFields);
         attributeName = sanitiseMetadataFieldName(filterType, attributeName);
         target[filterType][prefix + attributeName] = value;
       } else if (value) {
-        this.buildFormDetails(attributeName, target, value);
+        if (value[0]) { // is array and has index zero populated
```
paul-graham, line 80: _Question:_
Why do it this way rather than `if (Array.isArray(value) && value.length > 0)`, which is the standard pattern for checking something is an array of non-zero length?

adnanakgun: Applied your suggestion

paul-graham (approved):

src/cases/services/search-filter.service.ts
paul-graham: Thanks, @adnanakgun. Approved your PR.

Mo-Lala84 (dismissed):
approved

Mo-Lala84 (approved):
approved

RiteshHMCTS (commented):
approved

PR #1014, opened 2021-04-23
Media viewer page refresh amends

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-3552

### Change description ###
Resolve media viewer page refresh last document reload issue

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
+++ b/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
@@ -9,7 +9,7 @@ import { StoreModule } from '@ngrx/store';
 import { EffectsModule } from '@ngrx/effects';
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
 import { RouterTestingModule } from '@angular/router/testing';
-import { SessionStorageService } from 'src/app/services';
+import { SessionStorageService } from '../../services/session-storage/session-storage.service'
 
 const GATEWAY_DOCUMENT_URL = 'http://localhost:1234/documents';
 const REMOTE_DOCUMENT_URL = 'https://www.example.com/binary';
--- a/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
+++ b/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
@@ -9,6 +9,7 @@ import { StoreModule } from '@ngrx/store';
 import { EffectsModule } from '@ngrx/effects';
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
 import { RouterTestingModule } from '@angular/router/testing';
+import { SessionStorageService } from 'src/app/services';
 
 const GATEWAY_DOCUMENT_URL = 'http://localhost:1234/documents';
 const REMOTE_DOCUMENT_URL = 'https://www.example.com/binary';
@@ -23,6 +24,7 @@ describe('MediaViewerWrapperComponent', () => {
     let component: MediaViewerWrapperComponent;
     let fixture: ComponentFixture<MediaViewerWrapperComponent>;
     let windowService;
+    let sessionStorageService;
     let mockAppConfig: any;
     let featureToggleService;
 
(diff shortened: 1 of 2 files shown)


src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
```diff
@@ -9,6 +9,7 @@ import { StoreModule } from '@ngrx/store';
 import { EffectsModule } from '@ngrx/effects';
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
 import { RouterTestingModule } from '@angular/router/testing';
+import { SessionStorageService } from 'src/app/services';
```
paul-graham, line 12: Shouldn't import from `src/`.

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
```diff
@@ -2,6 +2,7 @@ import { Component, OnInit, ViewEncapsulation } from '@angular/core';
 import { WindowService } from '@hmcts/ccd-case-ui-toolkit';
 import { Observable } from 'rxjs';
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
+import { SessionStorageService } from 'src/app/services';
```
paul-graham, line 5: Shouldn't import from `src/`.

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
```diff
@@ -27,22 +28,32 @@ export class MediaViewerWrapperComponent implements OnInit {
     public constructor(
         private windowService: WindowService,
         private featureToggleService: FeatureToggleService,
+        private sessionStorageService: SessionStorageService
```
paul-graham, line 31: Can we make all of these `readonly`?

paul-graham (changes requested):
Some imports need updating and I've also requested a change on the constructor parameters. :)

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
```diff
@@ -27,22 +28,32 @@ export class MediaViewerWrapperComponent implements OnInit {
...
     async ngOnInit() {
 
 
         const localStorageMedia = this.windowService.getLocalStorage(MEDIA_VIEWER);
-        if (localStorageMedia) {
+        let sessionStorageMedia = this.sessionStorageService.getItem(MEDIA_VIEWER);
+
+        if (!sessionStorageMedia && localStorageMedia) {
+            // move media viewer item from local to session storage to allow page refresh on multiple media viewer tabs
+            sessionStorageMedia = localStorageMedia;
```
rayliangatos, line 43: should we replace the setLocalStorage to setSessionStorage in ccd-case-ui-toolkit for 'media-viewer-info', and here change the getLocalStorage to getSessionStorage, rather than do an override sessionStorageMedia <= localStorageMedia here?

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.spec.ts
paulhowes-HMCTS: Changed to relative path. Still waiting for Phil's path change to go in...

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
paulhowes-HMCTS: Changed to relative path. Still waiting for Phil's path change to go in...

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
paulhowes-HMCTS: Done.

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
paulhowes-HMCTS: Not quite sure what you're getting at here @rayliangatos. The local storage is set on the click event on the case documents tab of the case details page prior to loading a new tab. The file above deals with the new tab, and so needs to retrieve the value from local storage, move it to the session and then incorporate the results in the mv-media-viewer parameters which then display the media file. This allows multiple tabs to contain different files which don't get reset if the tab is refreshed.

paul-graham (approved):
Looks all good from my point of view. I'm approving on that basis. If there is anything to be changed from @rayliangatos's review, I'll re-review. 👍

src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
rayliangatos: so do you need both localStorage and sessionStorage to accomplish this requirement? Or SessionStorage is enough? As I see in the code it's finally using the SessionStorage only.

rayliangatos (approved):

PR #1059, opened 2021-05-15
Feature/EUI-3619 tasks and permission

rayliangatos:

https://tools.hmcts.net/jira/browse/EUI-3619

### Change description ###
* add mock api framework
* mytask/availabletasks
* manage link permission

```
[ ] Yes
[x] No
```


phillip-whitaker (approved):

udaydenduluri33 (approved):
LGTMr

api/workAllocation2/util.ts
```diff
@@ -155,64 +116,27 @@ export function prepareRoleApiRequest(locationId?: number): any {
...
+ * @param permissions The list of permissions the user holds.
+ * @return actionList:Action[] the list of total actions user holds.
  */
-export function getWATaskActions(permissions: TaskPermission[], view: ActionViews, isTaskAssigned: boolean): Action[] {
+export function getActionsByPermissions(view, permissions: []): Action[] {
   let actionList: Action[] = [];
   permissions.forEach(permission => {
     switch (permission) {
       case TaskPermission.MANAGE:
-        actionList = actionList.concat(getWATaskActionsForManage(view, isTaskAssigned));
```
connorpgpmcelroy, line 171: Why has isTaskAssigned been removed here? There needs to be a check that the task is assigned in order to get the correct actions. This looks like it will return the same for assigned and unassigned tasks for All Work. Because we're looping through the tasks, the check whether task is assigned needs to be internal to one of the methods in this util file.

connorpgpmcelroy (changes requested):

api/workAllocation2/util.ts
rayliangatos: We don't need to check whether the task is assigned(isTaskAssigned), as the new mock APIs will call different apis to assigned or unassigned tasks. I will give you a demo on how it works.

rayliangatos: As discuss as we are not sure how the API returns assigned/unassigned tasks, we will keep this as simple as we can for the permission check.

PR #1060, opened 2021-05-15
Feature/e UI 3805 task assignment confirm phillip whitaker

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-3805

### Change description ###

[New Design Screenshot](https://tools.hmcts.net/jira/secure/attachment/271075/EUI-3805.PNG)

Added the interface for the confirmation screen for EUI-3805. 

IMPORTANT NOTE: As part of this work I had to upgrade both govuk-frontend (minor) and govuk_frontend_toolkit (Majoy 8.* -> 9.*). This was because the designs stipulated that I used the latest GDS [Tags](https://design-system.service.gov.uk/components/tag/) UI element. The red version as part of the design was _not_ available in our current version. I have smoke tested the changes and all looks good. I have also checked the release notes and there does not seem to be major changes. 

```
[ ] Yes
[x] No (BUT... I needed upgrade front_end_uk_toolkit)
```


src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
```diff
@@ -0,0 +1,42 @@
...
+            <td class="govuk-table__cell">Joe Blogs</td>
+            <td class="govuk-table__cell">Protection</td>
+            <td class="govuk-table__cell">Taylor House</td>
+            <td class="govuk-table__cell">Review respondent interview</td>
+            <td class="govuk-table__cell">
+              <strong class="govuk-tag govuk-tag--red">
+                01/01/1971
+              </strong>
+            </td>
+            <td class="govuk-table__cell">Judge Birch</td>
```
rayliangatos, line 31: hardcoded?

udaydenduluri33: Yeah agree with @rayliangatos  can you use mock data here?

phillip-whitaker: I spoke on stand-up about this and it was agreed that I should do this. I can change it but wish people would make up their mind.

phillip-whitaker: After discussion today on stand-up it was agreed to do this and that a mock was needed at this stage.

paul-graham: Why does this need a mock? Isn't this just showing the user the very selection they made on the previous screen? Why is the data not simply being passed across by some straightforward means?

A mock makes no sense to me at all. Where does the need to call an API feature into this?

paul-graham: Moreover, you're making a _selection_ on the previous page, which suggests you're going to send that selection to an endpoint somewhere so it can subsequently be retrieved from that endpoint (i.e., the mock) but I can't imagine the WA API is going to offer an endpoint for a temporary selection, which means you'd have to implement the temporary store within MC instead... so what you _really_ want is that temporary store to be _implemented_, not mocked.

Architecturally, that seems like a strange thing to do when you can simply pass the details from one page to the next... or even have two modes on a single page ("select" and "confirm"). 🤦‍♂️

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    }).compileComponents();
+    fixture = TestBed.createComponent(WrapperComponent);
+    wrapper = fixture.componentInstance;
+    component = wrapper.appComponentRef;
+    component.verb = TaskActionType.Reassign;
+    router = TestBed.get(Router);
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
```
paul-graham, line 48: Can you rename this to `it('should give me 100% code coverage for no effort')`? 😂

paul-graham: In all seriousness, it might be a good idea to at least assert that the `verb` is being picked up and displayed correctly...

paul-graham (changes requested):
Just the one small request from me to add a slightly more meaningful unit test. I don't agree with the mock thing at all.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.spec.ts
phillip-whitaker: OKay, will do

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
paul-graham: Actually, how about this for a mock in the .ts file:
```
public mockTask: Task = {
  id: 'bob',
  case_id: '1234567890123456',
  caseName: 'Joe Bloggs',
  caseCategory: 'Protection',
  location: 'Taylor House',
  taskName: 'Review respondent interview',
  dueDate: new Date('1971-01-01T00:00:00'),
  actions: []
};

public mockCaseworker: Caseworker = {
  idamId: 'fred',
  firstName: 'Judge',
  lastName: 'Birch',
  email: 'judge.birch@hmcts.gov.uk',
  location: { id: 'TH', locationName: 'Taylor House', services: [] };
};
```

And then do the bindings in the template. Job done.

rayliangatos (approved):
Aggreed as this is a show feature only, and you could(have to) re-implement this base on the actual search person api in next story.

paul-graham (approved):
Thanks for adding a _vaguely_ meaningful unit test, @phillip-whitaker. 😁

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
```diff
@@ -0,0 +1,42 @@
...
+            <td class="govuk-table__cell"><a class="govuk-link fright" href="javascript:void(0)"
+                id="change__link" aria-label="Change PBA account numbers">Change</a></td>
```
rayliangatos, line 33: Change PBA account numbers?

phillip-whitaker: Oops

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
udaydenduluri33: Discussed in the call today that it was agreed hard coding is fine on Friday's call [Which I was not aware of]

udaydenduluri33 (approved):
LGTMp

PR #1167, opened 2021-07-14
bug-EUI-4261-redirect-uri

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EUI-4261

### Change description ###
Redirecting to Idam login page throws CORS error

```
[ ] Yes
[x] No
```


api/user/index.ts
```diff
@@ -4,6 +4,9 @@ import { getConfigValue } from '../configuration'
 import { CASE_SHARE_PERMISSIONS, SESSION_TIMEOUTS } from '../configuration/references'
 
 export async function getUserDetails(req, res: Response, next: NextFunction) {
+  if (!req.session || !req.session.passport || !req.session.passport.user) {
+    res.send({});
```
paul-graham, line 8: _Question:_
Should you `return` here? If you don't do so, the code below (line 11 onwards) will execute and, given you already know that there is _isn't_ a `req.session.passport.user`, you'll end up in the `catch` block... which probably isn't what you want.
```suggestion
    return res.send({});
```

paul-graham (changes requested):
I think the code is missing a `return`.

api/user/index.ts
RiteshHMCTS: yes Paul you are right (as always), I should return the object. 
Done the change
thanks

paul-graham: Definitely not always! 😂 

I see you've changed it... and added `.status(200)`. Shouldn't it be `.status(401)` since the user isn't authenticated?

paul-graham (approved):
Thank you for making the requested changes, @RiteshHMCTS, and for answering my questions. 🙂

api/user/index.ts
```diff
@@ -4,6 +4,9 @@ import { getConfigValue } from '../configuration'
 import { CASE_SHARE_PERMISSIONS, SESSION_TIMEOUTS } from '../configuration/references'
 
 export async function getUserDetails(req, res: Response, next: NextFunction) {
```
kuda1992, line 6: Can you add a return type function `getUserDetails()`

kuda1992 (changes requested):

api/user/index.ts
RiteshHMCTS: updated as mentioned

kuda1992 (dismissed):

paul-graham (approved):
It feels a bit squiffy returning a 200 status code when the user is unauthenticated so I'm not entirely sold on that bit of code... but I'm not going to hold this up as I'm assured it's being appropriately handled elsewhere.

api/user/index.ts
RiteshHMCTS: regarding 401 code instead of 200; if we return 401 then we get error in console as unauthorized (401) which we don't want as we are redirecting to to Idam login if we don't find user details.

connorpgpmcelroy (approved):
Changes make sense and CVE has been added. LGTM

paul-graham (approved):
Just an extra known audit issue since I approved it... so all good.

PR #1169, opened 2021-07-14
Remove pagination parameters

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-4285

### Change description ###
Removing pagination parameters from the request sent as it is not necessary (the data is already in the url)

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/api/test/pact/pact-tests/wa/searchTasks.spec.ts
+++ b/api/test/pact/pact-tests/wa/searchTasks.spec.ts
@@ -137,7 +137,7 @@ describe("Task management api, Search task", () => {
                 return configValues[prop];
             });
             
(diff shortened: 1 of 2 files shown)


udaydenduluri33 (approved):
@connorpgpmcelroy  I agree with this change because MVP is short lived feature. But, the other PR with release2 has a strategic changes. Thanks

kuda1992 (dismissed):
Looks fine but would have prefered creating a new searchRequest object without the `pagination_parameters` instead of mutating the existing object. Don't like the idea of deleting a property from an object.

udaydenduluri33 (approved):
I can see PACTs are successful now. @sreekanthpuligadda

PR #1172, opened 2021-07-15
EUI-2976: Strategic activity tracker

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2976 (this is the epic)

### Change description ###
Pointing at the toolkit from https://github.com/hmcts/ccd-case-ui-toolkit/pull/857. Also implemented the feature toggle mechanism for selecting the Activity Tracking mode.

```
[ ] Yes
[x] No
```


api/proxy.config.ts
```diff
@@ -1,113 +1,115 @@
...
-     * of the node layer.
-     */
-    // applyProxy(app, {
-    //     rewrite: true,
-    //     rewriteUrl: '',
-    //     source: '/workallocation',
-    //     target: 'http://localhost:8080',
-    // })
-}
+  /**
```
paulhowes-HMCTS, line 103: (Nitpick)
Do you think this will ever be used? If not or not in the near future I'd be tempted to remove it completely.

api/proxy.config.ts
```diff
@@ -1,113 +1,115 @@
...
-        rewrite: true,
-        rewriteUrl: '/activity',
-        source: [
-            '/activity',
-        ],
-        target: getConfigValue(SERVICES_CCD_COMPONENT_API_PATH),
-    })
+  applyProxy(app, {
+    rewrite: true,
+    rewriteUrl: '/socket.io',
```
paulhowes-HMCTS, line 12: (Nitpick)
Does the url have to be socket.io? Bit revealing.

charts/xui-webapp/values.preview.template.yaml
```diff
@@ -16,6 +16,8 @@ nodejs:
      HEALTH_WORK_ALLOCATION_TASK_API: http://wa-task-management-api-aat.service.core-compute-aat.internal/health
      HEALTH_ROLE_ASSIGNMENT_API: http://am-role-assignment-service-aat.service.core-compute-aat.internal/health
      HEALTH_CASE_CASEWORKER_API: http://rd-caseworker-ref-api-aat.service.core-compute-aat.internal/health
+     SERVICES_CCD_ACTIVITY_API_PATH: http://ccd-case-activity-api-pr-276.service.core-compute-preview.internal
```
paulhowes-HMCTS, line 19: Preview URL - so likely to disappear at some point.

package.json
```diff
@@ -69,7 +69,7 @@
     "@angular/platform-browser": "^7.2.13",
     "@angular/platform-browser-dynamic": "^7.2.13",
     "@angular/router": "^7.2.13",
-    "@hmcts/ccd-case-ui-toolkit": "4.1.0-rc.13",
+    "@hmcts/ccd-case-ui-toolkit": "v4.0.0-strategic-activity-tracker-gamma",
```
paulhowes-HMCTS, line 72: (Nitpick)
Thought we weren't using "v" anymore.

src/app/services/ccd-config/ccd-case.config.ts
```diff
@@ -64,6 +64,14 @@ export class AppConfig extends AbstractAppConfig {
     return this.config.remote_document_management_url;
   }
 
+  public getHrsUrl() {
```
paulhowes-HMCTS, line 67: Could do with a return type. Also, what is Hrs?

paulhowes-HMCTS (approved):
A few comments. No showstoppers, just finicky stuff and a couple of questions. Bravo on adding the semicolons at the end of the lines!

api/proxy.config.ts
paul-graham: It _doesn't_ have to be, but that's the default URL and I was having some difficulty with the socket.io configuration to get it to point to anything else relative. We can't use /activity either because it'll clash with the other proxy directly beneath. I'll have another look to see what we can do.

api/proxy.config.ts
paul-graham: I simply changed the indenting on this but it's a good question. @udaydenduluri33, any thoughts on this? Can I delete it rathe than it being commented out?

charts/xui-webapp/values.preview.template.yaml
paul-graham: Yes. This is purely for testing purposes. This should be removed prior to any merging, as discussed with @ernestman28. 😁

package.json
paul-graham: This is a tag-release version for testing purposes and will change once it's been QAed. I've always stuck a "v" at the front of my tags... am I not supposed to? 😬🤷‍♂️

src/app/services/ccd-config/ccd-case.config.ts
paul-graham: Absolutely no idea. This is here due to a breaking change that was introduced in the CCD Case UI Toolkit. Good call, though - I'll add a return type. Same for the one below. To be fair, it looks like they all need one! 🤦‍♂️

paul-graham: Done.

package.json
paulhowes-HMCTS: According to https://tools.hmcts.net/confluence/display/EUI/Release+process+effective+sprint+22.1 you don't need the v at the front. Anyhow, not a showstopper!

src/app/services/ccd-config/ccd-case.config.ts
paulhowes-HMCTS: Great, thanks.

api/proxy.config.ts
paul-graham: Had a good go at this but it's being a nuisance. I think I'll leave it for the time being seeing as it's currently working and you had it down as a _nitpick_ anyway. 😁

package.json
paul-graham: No "v" on the new one. 😁

Mo-Lala84:
Not required

PR #1177, opened 2021-07-16
EUI-3955: All work > Tasks tab > Assign task

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3955
https://tools.hmcts.net/jira/browse/EUI-4194

### Change description ###
Ensure correct verb ("assign") is used on task assignment screens and messages. Also ensure navigation back to "All work" on completing task reassignment or cancelling the process.

```
[ ] Yes
[x] No
```


src/work-allocation-2/containers/task-assignment/task-assignment-container.component.spec.ts
```diff
@@ -98,7 +98,9 @@ describe('TaskAssignmentContainerComponent', () => {
 
     wrapper.tasks = null;
     window.history.pushState({ returnUrl: 'mywork/list', showAssigneeColumn: false }, '', 'mywork/list');
```
rayliangatos, line 100: would this be my-work/list?

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.spec.ts
```diff
@@ -98,7 +98,9 @@ describe('TaskAssignmentContainerComponent', () => {
...
-    fixture.detectChanges();
+    // Deliberately defer fixture.detectChanges() call to each test, to allow overriding the ActivatedRoute snapshot
+    // data with a different verb ("Assign")
+    // fixture.detectChanges();
```
rayliangatos, line 103: could these be removed?

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.ts
```diff
@@ -47,8 +43,10 @@ export class TaskAssignmentContainerComponent implements OnInit, OnDestroy {
   private get returnUrl(): string {
     let url: string = '/work/my-work/list';
```
rayliangatos, line 44: this is still hard coded, is that correct?
how about doing a trick here?
if(verb ==='assign') {let url: string = '/work/all-work/list';}
else if(verb ==='reassign') {let url: string = '/work/my-work/list';}

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
```diff
@@ -29,6 +29,17 @@ export class TaskAssignmentConfirmComponent implements OnInit {
     private readonly messageService: InfoMessageCommService) {
   }
 
+  private get returnUrl(): string {
+    let url: string = '/work/my-work/list';
```
rayliangatos, line 33: this is still hardcoded, is that correct?

Daniel-Lam: @rayliangatos - that's correct. The default return URL is the "My work" screen and is used as a fall-back if there is no other return URL.

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.spec.ts
Daniel-Lam: @rayliangatos - interestingly, the tests don't seem to depend on this but, for the sake of consistency, I'll change it to `'my-work/list'`.

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.spec.ts
Daniel-Lam: @rayliangatos - happy to remove Line 103 but I think I'll leave the comment in place to explain why `fixture.detectChanges()` isn't being called here (which it would usually).

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.ts
Daniel-Lam: @rayliangatos - similar to above, it's hard-coded because the default return URL is the "My work" screen. It's not possible to set it based on the verb because, having discussed this with Tommy Wong (BA), it turns out the "All work" screen has both "assign" and "reassign" actions on tasks.

rayliangatos: I double check the code, actually this default value is useless, and it will be always overrided by:
`    if (window && window.history && window.history.state && window.history.state.returnUrl) {
      url = window.history.state.returnUrl;
    }`

In that case could you assign the empty value to it? like:
```suggestion
    let url: string = '';
```
Then it will be a genuine generic component.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
rayliangatos: I double check the code, actually this default value is useless, and it will be always overrided by:
if (window && window.history && window.history.state && window.history.state.returnUrl) { url = window.history.state.returnUrl; }

In that case could you assign the empty value to it? like:

rayliangatos (changes requested):
request change comment added

src/work-allocation-2/containers/task-assignment/task-assignment-container.component.ts
Daniel-Lam: Not always, @rayliangatos... I came across a case where `returnUrl` is `undefined` if the user uses the browser navigation buttons to go back and then forwards - hence I had to check for `window.history.state.returnUrl`.

rayliangatos: We can't control user and prohibit them to use the back button in browser, but in that case if user use the Back button in browser when in All Work->tasks, this cancel button will go back to My Work->my tasks, that shouldn't be allowed, we would rather let the user go back to their home page, that is '' for them. As not everyone can access both My Work and ALL Work. I would give you a demo but hard code as '/work/my-work/list' this cause an edge case issue in All Work.

Daniel-Lam: Ah... I didn't know that - I thought everyone who had access to "All work" also had "My work" but this appears not to be the case. I will default it to `''`, then. I think there is also another location where I will have to change it to `''`.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
Daniel-Lam: Similar to above, I will change the default return URL to `''`.

rayliangatos (approved):

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
```diff
@@ -1,13 +1,13 @@
...
     <span class="govuk-caption-l">{{verb}} task</span>
-    Check your changes
+    Check your answers
   </h1>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
 
       <div id="reassign-confirm-hint" class="govuk-hint">
-        Check you are reassigning work to the right person.
+        Check you are {{verb | lowercase }}ing work to the right person.
```
paulhowes-HMCTS, line 10: Not a big fan of 'calculating' words - plays havoc when it comes round to translating phrases

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
```diff
@@ -58,11 +71,12 @@ export class TaskAssignmentConfirmComponent implements OnInit {
...
 
   public onCancel(): void {
-    this.router.navigate([this.rootPath, 'my-work', 'list']);
+    // Use returnUrl to return the user to the "All work" or "My work" screen, depending on which one they started from
+    this.router.navigate([this.returnUrl]);
   }
 
   private reportSuccessAndReturn(): void {
-    const message = InfoMessage.REASSIGNED_TASK;
+    const message = this.verb === 'Assign' ? InfoMessage.ASSIGNED_TASK : InfoMessage.REASSIGNED_TASK;
```
paulhowes-HMCTS, line 79: Again, calculating verbs makes me feel queasy.

paulhowes-HMCTS (changes requested):
Just one point about calculating verbs - would cause issues if translating (i.e. to Welsh).

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
Daniel-Lam: @paulhowes-HMCTS - hmm... good point. Happy to look at an alternative approach if you have one.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
Daniel-Lam: @paulhowes-HMCTS - as above, happy to look at an alternative approach.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.html
paulhowes-HMCTS: Maybe just stick the verb logic in the code file and choose the relevant phrase as opposed to just the word.

Daniel-Lam: @paulhowes-HMCTS - I've reimplemented this as suggested; please re-review - thanks.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
Daniel-Lam: @paulhowes-HMCTS - are you happy to resolve this conversation, given we discussed that here the verb is used as a discriminator to select the correct message from the `InfoMessage` enum?

paulhowes-HMCTS (approved):
Approved now the verb calculation has been amended.

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.ts
paulhowes-HMCTS: Resolved!

PR #1189, opened 2021-07-21
EUI-3777 create prepended tabs on case view v2

kuda1992:

https://tools.hmcts.net/jira/browse/EUI-3777

### Change description ###
Moved the caseviewer route from toolkit to MC
Added to extra container components which will house the tasks and roles and access tab.

```
[X] Yes
[ ] No
```


src/app/services/ccd-config/ccd-case.config.ts
```diff
@@ -148,4 +158,38 @@ export class AppConfig extends AbstractAppConfig {
...
+        (isFeatureEnabled) =>
+          this.workallocationUrl = AppUtils.getFeatureToggledUrl(
+            isFeatureEnabled,
+            WorkAllocationTaskService.WorkAllocationUrl
+          )
+      );
+  }
+
+  private enablePrependedTabs(feature: string, userDetails: UserDetails): boolean {
+    return feature === 'WorkAllocationRelease2'
```
udaydenduluri33, line 192: Please refer constants and string literal

src/cases/case-feature.routes.ts
```diff
@@ -102,7 +105,26 @@ export const ROUTES: Routes = [
...
+                {
+                  path: '',
+                  pathMatch: 'full',
+                },
+                {
+                  path: 'tasks',
+                  component: TasksContainerComponent
+                },
+                {
+                  path: 'roles and access',
```
udaydenduluri33, line 122: remove space in the path

src/app/services/ccd-config/ccd-case.config.ts
```diff
@@ -148,4 +158,38 @@ export class AppConfig extends AbstractAppConfig {
...
+    this.featureToggleService
+      .isEnabled(AppConstants.FEATURE_NAMES.workAllocation)
+      .subscribe(
+        (isFeatureEnabled) =>
+          this.workallocationUrl = AppUtils.getFeatureToggledUrl(
+            isFeatureEnabled,
+            WorkAllocationTaskService.WorkAllocationUrl
+          )
+      );
+  }
```
rayliangatos, line 189: Could these move to the container(exui container) level? 
The problem to put here is ccd-case.config.ts is intantiated/implemented when the APP is started, we shouldn't put the container's dynamic logic here.
The ccd-case.config.ts is used for put all those configuration(constant immutable value) settings.

rayliangatos (changes requested):
leave request change comment

src/app/services/ccd-config/ccd-case.config.ts
kuda1992: @rayliangatos I cannot move this to container level. As this config needs to be passed to the `toolkit`. As well as its according to the design. 
https://tools.hmcts.net/confluence/display/EUI/CCD+Toolkit+-+Custom+added+tabs

rayliangatos (approved):
As discussed, we will probably refactor the current design in next IP week, for timebeing approve the PR.

connorpgpmcelroy (approved):
Can see comments have been addresses and no issues with code seen from me

PR #1207, opened 2021-07-27
Feature/exlucsion integration 2

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-3946

### Change description ###
Integration of add exclusion journey, refactoring the folder path/route.

```
[ ] Yes
[x] No
```


src/work-allocation-2/components/find-person/find-person.component.ts
```diff
@@ -14,10 +16,10 @@ import { FindAPersonService } from '../../services/find-person.service';
 export class FindPersonComponent implements OnInit {
   @Output() public personSelected = new EventEmitter<Person>();
   @Input() public title: string;
-  @Input() public domainString: string = 'BOTH';
+  @Input() public domainString: string = 'All';
```
connorpgpmcelroy, line 19: As mentioned on Slack, giving the input as a string was causing issues in common-lib, could we not pass in the enum?

src/work-allocation-2/components/find-person/find-person.component.ts
```diff
@@ -44,7 +46,7 @@ export class FindPersonComponent implements OnInit {
   }
 
   public filter(searchTerm: string): Observable<Person[]> {
-    if (searchTerm && searchTerm.length > this.minSearchCharacters) {
+    if (searchTerm && searchTerm.length >= this.minSearchCharacters) {
```
connorpgpmcelroy, line 49: This may be why 2 characters start the search instead of 3 - If we want >= then we need to reset minSearchCharacters to 3

src/work-allocation-2/components/find-person/find-person.component.ts
```diff
@@ -59,7 +61,7 @@ export class FindPersonComponent implements OnInit {
   }
 
   public updatedVal(currentValue) {
-    if (currentValue && currentValue.length > this.minSearchCharacters) {
+    if (currentValue && currentValue.length >= this.minSearchCharacters) {
```
connorpgpmcelroy, line 64: Same as above

connorpgpmcelroy (changes requested):
Main point of contention is with the minSearchCharacters and how that will affect the findPerson calls. I also think we should use enums for passing into the findPerson component instead of domainString but will have call with Ray to address this

src/work-allocation-2/components/find-person/find-person.component.ts
rayliangatos: I now set the this.minSearchCharacters=3, that make it more sense for minSearchCharacters, this code will keep as >=

src/work-allocation-2/components/find-person/find-person.component.ts
rayliangatos: I now set the this.minSearchCharacters=3, that make it more sense for minSearchCharacters, this code will keep as >=

src/work-allocation-2/components/find-person/find-person.component.ts
rayliangatos: as show on demo

connorpgpmcelroy (approved):
All looks good and need these changes to ensure other changes are correct. minSearchCharacters has been changed as per Andy and Tommy's recommendation

api/workAllocation2/constants/mock.data.ts
```diff
@@ -1,61 +1,70 @@
...
+    domain: 'JUDICIAL'
   },
   {
-    id: 'id124',
-    name: 'Andy Wilkins',
-    email: 'andy.wilkins@email.com',
-    domain: PersonDomain.JUDICIAL
+    id: 'id103',
+    name: 'James Priest',
+    email: 'james.priest@judicail.com',
```
paulhowes-HMCTS, line 24: Misspelling of judicial

api/workAllocation2/util.ts
```diff
@@ -197,8 +197,8 @@ export function getActionsByPermissions(view, permissions: TaskPermission[]): Ac
   return actionList;
 }
 
-export function applySearchFilter(person: Person, domain: PersonDomain, searchTerm: any): boolean {
-  if (domain === PersonDomain.BOTH) {
+export function applySearchFilter(person: Person, domain: string, searchTerm: any): boolean {
+  if (domain === 'ALL') {
```
paulhowes-HMCTS, line 201: Why is this being changed to a string as opposed to keeping the enumerable value?

api/workAllocation2/utils.spec.ts
```diff
@@ -270,29 +276,34 @@ describe('workAllocation.utils', () => {
...
+        .to.deep.equal([ASSIGN, GO, COMPLETE]);
     });
 
   });
 
   describe('applySearchFilter', () => {
     it('PersonDomain BOTH', () => {
       const person = {id: '123', name: 'some name', email: 'name@email.com', domain: PersonDomain.CASEWORKER };
-      const result = applySearchFilter(person, PersonDomain.BOTH, 'name');
+      const result = applySearchFilter(person, 'ALL', 'name');
```
paulhowes-HMCTS, line 306: I prefer PersonDomain.ALL

src/work-allocation-2/components/find-person/find-person.component.ts
paulhowes-HMCTS: Can't we use PersonDomain.ALL?

src/work-allocation-2/components/find-person/find-person.component.ts
```diff
@@ -14,26 +16,26 @@ import { FindAPersonService } from '../../services/find-person.service';
 export class FindPersonComponent implements OnInit {
   @Output() public personSelected = new EventEmitter<Person>();
   @Input() public title: string;
-  @Input() public domainString: string = 'BOTH';
+  @Input() public domainString: string = 'All';
   @Input() public findPersonGroup;
   @Input() public selectedPerson: string;
-  public domain: PersonDomain;
+  public domain: string;
```
paulhowes-HMCTS, line 22: Can you explain why it is now a string please?

paulhowes-HMCTS (changes requested):
A few things:
Ticket description needs fixing
Misspelling of 'exlucsion' in branch title, plus a typo as per comments
Do numerous empty css files still exist? If they are not being used and we are not referencing them in the template they can be deleted

I also have a query as to why you are changing the logic to use string comparisons as opposed to enumerables. This would result in having to change code in multiple areas both in the template and code files and makes the bindings 'less strict'.

api/workAllocation2/util.ts
rayliangatos: the old PersonDomain is with
JUDICIAL = 1,
CASEWORKER = 2,
BOTH = 3,

We can not pass 1, 2, 3 from the container and it's not meaningful, we can still use the string enum, but giving that there is no difference on using string enum, so a common const string is almost the same.

api/workAllocation2/utils.spec.ts
rayliangatos: explained on above

src/work-allocation-2/components/find-person/find-person.component.ts
rayliangatos: as explained above

api/workAllocation2/constants/mock.data.ts
rayliangatos: changed to judicial

rayliangatos:
> A few things:
> Ticket description needs fixing - **Updated**
> Misspelling of 'exlucsion' in branch title, plus a typo as per comments - **Good shout but I can't change the branch title as it needs re-raise PR, ask re-approval etc. And this doesn't matter on branch name when merging stuff in.**
> Do numerous empty css files still exist? If they are not being used and we are not referencing them in the template they can be deleted - **Empty css files have been removed.**
> I also have a query as to why you are changing the logic to use string comparisons as opposed to enumerables. This would result in having to change code in multiple areas both in the template and code files and makes the bindings 'less strict'.
- **I have replied with the comment.**

api/workAllocation2/util.ts
paulhowes-HMCTS: I don't think we should be using 'ALL', we should be using the enumerated string.

api/workAllocation2/utils.spec.ts
paulhowes-HMCTS: Let's discuss.

api/workAllocation2/util.ts
rayliangatos: it's now enum, could you double check

api/workAllocation2/utils.spec.ts
rayliangatos: it's now enum, could you double check?

paulhowes-HMCTS (approved):
Changes have been made based on comments so approved.

PR #1211, opened 2021-07-28
remove the busines logic from ccd-config

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-4324

### Change description ###
Move the prepended tabs business logic from the cdd config.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
+++ b/src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
@@ -1,11 +1,13 @@
 import { Component, Input } from '@angular/core';
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { async, ComponentFixture, fakeAsync, TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
-import { CaseView, CaseTab } from '@hmcts/ccd-case-ui-toolkit';
-import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
+import { CaseTab, CaseView } from '@hmcts/ccd-case-ui-toolkit';
+import { FeatureToggleService, FeatureUser } from '@hmcts/rpx-xui-common-lib';
+import { StoreModule } from '@ngrx/store';
+import { provideMockStore } from '@ngrx/store/testing';
+import { Observable, of } from 'rxjs';
+import { reducers, State } from 'src/app/store';
 import { CaseViewerContainerComponent } from './case-viewer-container.component';
-import {StoreModule} from '@ngrx/store';
-
 
 @Component({
   // tslint:disable-next-line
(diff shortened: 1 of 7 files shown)


leepavlouhmcts (approved):

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
```diff
@@ -1,20 +1,63 @@
...
   public caseDetails: CaseView;
+  public tabs$: Observable<CaseTab[]>;
 
-  constructor(private readonly route: ActivatedRoute) {
+  constructor(private readonly route: ActivatedRoute,
+              private readonly store: Store<fromRoot.State>,
+              private readonly featureToggleService: FeatureToggleService) {
   }
 
   public ngOnInit() {
```
connorpgpmcelroy, line 44: : void

connorpgpmcelroy (changes requested):
All looks good apart from ngOnInit missing return type

src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
```diff
@@ -32,7 +104,12 @@ describe('CaseViewerContainerComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should create', () => {
-    expect(component).toBeTruthy();
+  it('should return the two tabs', (done: DoneFn) => {
```
paul-graham, line 107: Is there a test on the _actual_ `ccd-case-viewer` component to make sure it renders the tabs it receives? This doesn't test that anything within the HTML is wired up even with the mock `ccd-case-viewer` and I think it could benefit from at least testing that...

src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
```diff
@@ -1,29 +1,101 @@
 import { Component, Input } from '@angular/core';
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { async, ComponentFixture, fakeAsync, TestBed } from '@angular/core/testing';
```
paul-graham, line 2: _Nitpick:_
You've introduced `fakeAsync` but it doesn't look like it's being used anywhere in here.

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
```diff
@@ -1,20 +1,63 @@
...
+import * as fromRoot from '../../../app/store';
 
 @Component({
   selector: 'exui-case-viewer-container',
   templateUrl: './case-viewer-container.component.html',
   styleUrls: ['./case-viewer-container.component.scss']
 })
 export class CaseViewerContainerComponent implements OnInit {
+  private static WORKALLOCATIONRELEASE1 = 'WorkAllocationRelease1';
+  private static WORKALLOCATIONRELEASE2 = 'WorkAllocationRelease2';
```
paul-graham, line 20: _Nitpick:_
Subject for a different conversation and I know it's simply been moved, but... ugh! 🤢

paul-graham (changes requested):
I think the tests are incomplete. What's being tested right now is beneficial as it checks that the wiring up within the `ngOnInit()` works, but there's nothing to test that the wiring up within the HTML is appropriate.

src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
kuda1992: `ccd-case-viewer` component is part of toolkit. Toolkit already tests if the component renders the tabs.

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
kuda1992: What do you suggest here @paul-graham ?

src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
paul-graham: Okay, cool. This still doesn't test that you've wired up the HTML, though. 😁

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
paul-graham: Well, it's kind of awful that they're there in the first place - they should be declared somewhere that _anywhere_ that needs to toggle on the version of Work Allocation being used can get at them. In addition, there's no casing and they're also mutable so not really constants. I'd go with something more like:
```
public static readonly FEATURE_WORK_ALLOCATION_RELEASE_1 = 'WorkAllocationRelease1';
public static readonly FEATURE_WORK_ALLOCATION_RELEASE_2 = 'WorkAllocationRelease2';
```

And I've put them as `public` because I think they should be elsewhere. I'm assuming `'WorkAllocationRelease1` and `WorkAllocationRelease2` are LaunchDarkly values or something.

kuda1992: I am going to leave them as private because there aren't being used anywhere else. However they're there because we are checking which feature is activated in launch darkly. If its WorkAllocationRelease2 the pretended tabs should be shown.

src/cases/containers/case-viewer-container/case-viewer-container.component.spec.ts
paul-graham: Ah-ha! You've added another test beneath without even telling me. A nice surprise as I was just looking for the things you've _said_ you've addressed! 😁

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
paul-graham: Ah, you've got to love a **pretend** tag. 😂 

No problem, with respect to them remaining `private` if this is really the only place those constants are needed. Thanks for at least giving them better names. 🙂

paul-graham (approved):
Thank you for making the requested changes, @kuda1992, and explaining the context of some other things I commented on. 👍

connorpgpmcelroy (approved):
All looks good and comment addressed

PR #1213, opened 2021-07-28
EUI-4170: Roles and access > Delete exclusion > "Check your answers" page

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-4170

### Change description ###
Implement display-only User Story.

```
[ ] Yes
[x] No
```


src/role-access/containers/delete-exclusion/delete-exclusion.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(private readonly route: ActivatedRoute) {}
+
+  public ngOnInit(): void {
+    // Get the role exclusions from the route, which will have been put there by the resolver.
+    const roleExclusions = this.route.snapshot.data.roleExclusions;
+    this.populateAnswers(roleExclusions);
+  }
+
+  public onEventTrigger(event: ExclusionNavigationEvent) {
```
connorpgpmcelroy, line 29: Method has no void return type

src/role-access/containers/delete-exclusion/delete-exclusion.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.eventTrigger.emit(event);
+  }
+
+  private populateAnswers(exclusions: any[]) {
```
connorpgpmcelroy, line 33: Could any[] change to be Exclusions or similar? Void return type for this as well?

connorpgpmcelroy (changes requested):
Just two minor comments

rayliangatos (approved):

src/role-access/containers/delete-exclusion/delete-exclusion.component.ts
Daniel-Lam: @connorpgpmcelroy - good catch! Will add `void`

src/role-access/containers/delete-exclusion/delete-exclusion.component.ts
Daniel-Lam: @connorpgpmcelroy - you're right... I'll fix both of these

connorpgpmcelroy (approved):
Have checked throughout and all looks good

PR #1244, opened 2021-08-16
allocate link directive

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-3799

### Change description ###
Created a new endpoint which takes the caseId and checks if the user can allocate a role

```
[ ] Yes
[x] No
```

### DIFF
--- a/api/workAllocation2/constants/roles.mock.data.ts
+++ b/api/workAllocation2/constants/roles.mock.data.ts
@@ -1,4 +1,4 @@
-import { CaseRole } from '../interfaces/caseRole';
+import { CaseRole, RoleType } from '../interfaces/caseRole';
 
 export const CASEROLES: CaseRole[] = [
   {
@@ -10,7 +10,7 @@ export const CASEROLES: CaseRole[] = [
     id: 'd90ae606-98e8-47f8-b53c-a7ab77fde22b',
     location: 'Taylor House',
     name: 'Judge Beech',
-    role: 'Lead Judge',
+    role: RoleType.LEAD_JUDGE,
     start: '2021-07-13T00:29:10.656Z',
   },
   {
@@ -22,7 +22,7 @@ export const CASEROLES: CaseRole[] = [
     id: 'd90ah606-98e8-47f8-b53c-a7ab77fde22b',
     location: 'Milton Keynes',
     name: 'Kuda Nyamainashe',
-    role: 'Lead Judge',
+    role: RoleType.LEAD_JUDGE,
     start: '2021-05-19T00:29:10.656Z',
   },
 ];
--- a/api/workAllocation2/constants/roles.mock.data.ts
+++ b/api/workAllocation2/constants/roles.mock.data.ts
@@ -1,4 +1,5 @@
-import { CaseRole, RoleType } from '../interfaces/caseRole';
+import { TypeOfRole } from '../../../src/role-access/models';
+import { CaseRole } from '../interfaces/caseRole';
 
 export const CASEROLES: CaseRole[] = [
   {
@@ -10,7 +11,7 @@ export const CASEROLES: CaseRole[] = [
     id: 'd90ae606-98e8-47f8-b53c-a7ab77fde22b',
     location: 'Taylor House',
     name: 'Judge Beech',
-    role: RoleType.LEAD_JUDGE,
+    role: TypeOfRole.LEAD_JUDGE,
     start: '2021-07-13T00:29:10.656Z',
   },
   {
@@ -22,7 +23,7 @@ export const CASEROLES: CaseRole[] = [
     id: 'd90ah606-98e8-47f8-b53c-a7ab77fde22b',
     location: 'Milton Keynes',
     name: 'Kuda Nyamainashe',
-    role: RoleType.LEAD_JUDGE,
+    role: TypeOfRole.LEAD_JUDGE,
     start: '2021-05-19T00:29:10.656Z',
   },
 ];
--- a/api/workAllocation2/interfaces/caseRole.ts
+++ b/api/workAllocation2/interfaces/caseRole.ts
@@ -1,6 +1,8 @@
+import { TypeOfRole } from '../../../src/role-access/models';
+
 export interface CaseRole {
   name: string;
-  role: RoleType;
+  role: TypeOfRole;
   location: string;
   start: string;
   end: string;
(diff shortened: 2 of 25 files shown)


api/workAllocation2/roleService.ts
```diff
@@ -12,3 +13,15 @@ export async function handleGetRolesByCaseId(path: string, req: EnhancedRequest)
   const headers = setHeaders(req);
   return await httpMock.get<CaseRole>(path, {headers});
 }
+
+export function handleShowAllocatorLinkByCaseId(caseId: string, req: EnhancedRequest): boolean {
```
udaydenduluri33, line 17: @kuda1992  I can see you looking for caseId below. But, you also need to look if the the user has the case allocator role. That is missing. Could you please add it? Thanks

api/workAllocation2/interfaces/caseRole.ts
```diff
@@ -12,3 +12,10 @@ export interface Action {
   id: string;
   title: string;
 }
+
+export enum RoleType {
+  LEAD_JUDGE = 'Lead Judge',
```
rayliangatos, line 17: there has been a similar TypeOfRole in src\role-access\models\allocate-role.enum

src/cases/directives/show-allocate-link.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+  @Input() public showAllocateRoleLink: boolean = false;
+  @Input() public roles: CaseRole[];
+  @Input() public userType: string;
+
+  constructor(private element: ElementRef<HTMLElement>) {
+  }
+
+  private static canDisplayLink(roles: CaseRole[], userType: string, showAllocateRoleLink: boolean): boolean {
+    const show = false;
+    if (!showAllocateRoleLink || userType === 'legalOps' && ShowAllocateLinkDirective.hasExceededNumberCaseManagerRoles(roles)) {
```
connorpgpmcelroy, line 18: 'legalOps' should be a constant/enum along with 'judiciary' so if we need to change any mention of it, looking through code is not necessary

connorpgpmcelroy (changes requested):
Further to Uday's comment, would prefer legalops and judiciary in a constants/enum file

src/cases/directives/show-allocate-link.directive.ts
```diff
@@ -0,0 +1,37 @@
+import { Directive, ElementRef, Input, OnInit } from '@angular/core';
+import { CaseRole } from 'api/workAllocation2/interfaces/caseRole';
+import { TypeOfRole } from 'src/role-access/models';
```
udaydenduluri33, line 3: please move src to relative path

src/cases/components/roles-and-access/roles-and-access.component.ts
```diff
@@ -1,13 +1,17 @@
 import { Component, Input } from '@angular/core';
 import { CaseView } from '@hmcts/ccd-case-ui-toolkit';
-import { CaseRole } from '../../../../api/workAllocation2/interfaces/caseRole';
 import { LocationInfo } from 'src/app/store/reducers/app-config.reducer';
```
udaydenduluri33, line 3: please move src to relative path

src/cases/directives/show-allocate-link.directive.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { TypeOfRole } from 'src/role-access/models';
```
udaydenduluri33, line 1: please move src to relative path

src/cases/resolvers/case-roles-resolver.service.spec.ts
```diff
@@ -4,17 +4,15 @@ import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
 import { ActivatedRouteSnapshot } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
-import { StoreModule } from '@ngrx/store';
+import { of } from 'rxjs';
+import { TypeOfRole } from 'src/role-access/models';
```
udaydenduluri33, line 8: please move src to relative path

api/workAllocation2/constants/roles.mock.data.ts
```diff
@@ -1,3 +1,4 @@
+import { TypeOfRole } from '../../../src/role-access/models';
```
connorpgpmcelroy, line 1: import contains 'src'

api/workAllocation2/util.ts
```diff
@@ -206,8 +206,12 @@ export function applySearchFilter(person: Person, domain: string, searchTerm: an
...
   const headers = setHeaders(req);
-  const response: AxiosResponse = await http.post(path, payload, { headers });
+  const response: AxiosResponse = await http.post(path, payload, {headers});
   // Return the whole response, not just the data, so we can
   // see what the status of the response is.
   return response;
 }
+
+export function hasCaseAllocatorRole(authorisations: string[]): boolean {
+  return authorisations.some(authorisation => authorisation === 'case-allocator-role');
```
connorpgpmcelroy, line 216: Think 'case-allocator-role' may already be a constant within the application so we should use that

src/role-access/components/case-roles-table/case-roles-table.component.html
```diff
@@ -36,7 +36,7 @@
 <dl class="govuk-summary-list" *ngIf="!roles ||  !roles.length">
   <div class="govuk-summary-list__row">
     <dd class="govuk-summary-list__value">
-      There are no {{roleType}} roles for this case.
+      There are no {{userType === 'legalOps' ? 'legal ops' : userType }} roles for this case.
```
connorpgpmcelroy, line 39: Could we not use the enum here?

connorpgpmcelroy (changes requested):
Minor changes required

api/workAllocation2/util.ts
connorpgpmcelroy: checked and it's a constant in app.constants.ts

connorpgpmcelroy: and also in api/user/constants.ts

udaydenduluri33 (approved):
Approving after the changes requested

connorpgpmcelroy (approved):
Comments have been discussed/addressed and functionality looks good

PR #1255, opened 2021-08-19
Feature/e UI 4069 choose role angular changes

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-4069

### Change description ###
Creating service to check for judicial and legalOps roles and storing them until they are used in either allocate role user story.

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/api/workAllocation2/constants/roles.mock.data.ts
+++ b/api/workAllocation2/constants/roles.mock.data.ts
@@ -38,7 +38,7 @@ export const CASEROLES: CaseRole[] = [
     id: 'd90ah606-98e8-47r8-b53c-a7ab77fde22b',
     location: 'Taylor House',
     name: 'Bisa Bulter',
-    role: TypeOfRole.CASE_MANAGER,
+    role: TypeOfRole.CaseManager,
     start: '2021-08-19T00:29:10.656Z',
   },
 ];
--- a/src/cases/components/roles-and-access/roles-and-access.component.ts
+++ b/src/cases/components/roles-and-access/roles-and-access.component.ts
@@ -30,12 +30,12 @@ export class RolesAndAccessComponent implements OnInit {
   @Input()
   public set roles(value: CaseRole[]) {
     this.pRoles = value;
-    this.legalOpsRoles = this.roles.filter(role => role.role === TypeOfRole.CASE_MANAGER);
+    this.legalOpsRoles = this.roles.filter(role => role.role === TypeOfRole.CaseManager);
     this.judicialRoles = this.roles.filter(role => RolesAndAccessComponent.isJudicialRole(role));
   }
 
   public static isJudicialRole(caseRole: CaseRole): boolean {
-    return caseRole.role === TypeOfRole.LEAD_JUDGE || caseRole.role === TypeOfRole.HEARING_JUDGE;
+    return caseRole.role === TypeOfRole.LeadJudge || caseRole.role === TypeOfRole.HearingJudge;
   }
 
   public ngOnInit(): void {


api/roleAssignment/roleAssignmentService.ts
```diff
@@ -0,0 +1,15 @@
+import { NextFunction, Response } from 'express';
+import { EnhancedRequest } from '../lib/models';
+
+export async function getJudicialRoles(req: EnhancedRequest, res: Response, next: NextFunction) {
```
kuda1992, line 4: **Minor** can we have a return type.

api/roleAssignment/roleAssignmentService.ts
```diff
@@ -0,0 +1,15 @@
...
+  const personRoles = [
+    {roleId: 'lead', roleName: 'Lead judge'},
+    {roleId: 'hearing', roleName: 'Hearing judge'}];
+  return res.send(personRoles).status(200);
+}
+
+export async function getLegalOpsRoles(req: EnhancedRequest, res: Response, next: NextFunction) {
```
kuda1992, line 11: Same here as well return type

api/routes.ts
```diff
@@ -27,6 +28,8 @@ router.use('/user', userRouter);
 
 router.use('/exclusions', exclusionRouter);
 
+router.use('/role-assignment', roleAssignmentRouter)
```
kuda1992, line 31: missing semi-colon 😊

src/role-access/containers/allocate-role/choose-role/choose-role.component.spec.ts
```diff
@@ -1,63 +1,103 @@
...
-    spyOnPipeToStore = spyOn(store, 'pipe').and.callThrough();
-    spyOnStoreDispatch = spyOn(store, 'dispatch');
     fixture = TestBed.createComponent(ChooseRoleComponent);
     component = fixture.componentInstance;
     component.formGroup = formGroup;
+    mockStore.pipe.and.returnValue(of(mockAllocateRoleStateData));
     fixture.detectChanges();
   });
 
   it('should create', () => {
```
kuda1992, line 84: Might as well remove this test now that you have added other tests.

src/role-access/containers/allocate-role/choose-role/choose-role.component.ts
```diff
@@ -91,6 +81,18 @@ export class ChooseRoleComponent implements OnInit, OnDestroy {
     }
   }
 
+  private getOptions(roles: Role[]): OptionsModel[] {
+    const options = [];
```
kuda1992, line 85: could simplify this code 
from
`
roles.forEach((jr) => {
      const thisOption: OptionsModel = {
        optionId: jr.roleId,
        optionValue: jr.roleName
      };
      options.push(thisOption);
    });
    return options;`
to 

`return roles.map(jr => ({optionId: jr.roleId, optionValue: jr.roleName}))`;

src/role-access/services/role-assignment.service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class RoleAssignmentService {
+  public static allocationsUrl = '/api/role-assignment';
+
+  public judicialRoles: Role[];
+  public legalOpsRoles: Role[];
+  constructor(private readonly http: HttpClient) { }
+
+  public setRoleAllocations(): void {
```
kuda1992, line 15: Could use `forkJoin` and avoid multiple subscribes. However this doesn't look right why is the service method subscribing to itself.

kuda1992 (changes requested):
Mostly good left a fe comments

src/role-access/services/role-assignment.service.ts
connorpgpmcelroy: Done

src/role-access/containers/allocate-role/choose-role/choose-role.component.ts
connorpgpmcelroy: Done

src/role-access/containers/allocate-role/choose-role/choose-role.component.spec.ts
connorpgpmcelroy: Done

api/routes.ts
connorpgpmcelroy: Done

api/roleAssignment/roleAssignmentService.ts
connorpgpmcelroy: Done

api/roleAssignment/roleAssignmentService.ts
connorpgpmcelroy: Done

kuda1992 (approved):
Thanks for making the changes @connorpgpmcelroy

rayliangatos (changes requested):

api/routes.ts
```diff
@@ -27,6 +28,8 @@ router.use('/user', userRouter);
 
 router.use('/role-access', roleAccessRouter);
 
+router.use('/role-assignment', roleAssignmentRouter);
```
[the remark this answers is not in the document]
connorpgpmcelroy: Yes, can do this. This was a change made after I originally implemented this and the ticket focuses on the Angular layer but can rearrange it

src/app/containers/app/app.component.ts
```diff
@@ -81,6 +88,16 @@ export class AppComponent implements OnInit, OnDestroy {
     }
   }
 
+  // EUI-4069 - Get roles to display on allocate role journey one time on login
+  public setRoleAllocations(): void {
```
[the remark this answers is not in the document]
connorpgpmcelroy: 1. It doesn't but was requested by Kuda. If you think it would be more appropriate to split these calls then maybe we should have a call and include Kuda?
2. When discussing with Andy he mentioned we should not be making multiple calls for this as the roles are unlikely to change. take(1) ensures that we only get the initial returned result.
Do you still want me to change it?

src/role-access/containers/allocate-role/choose-role/choose-role.component.ts
```diff
@@ -8,9 +8,10 @@ import { $enum as EnumUtil } from 'ts-enum-util';
 import { UserRole } from '../../../../app/models/user-details.model';
 import * as fromAppStore from '../../../../app/store';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Will remove

src/role-access/services/role-assignment.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Injectable } from '@angular/core';
+import { Role } from '../models/role.model';
+
+@Injectable({ providedIn: 'root' })
+export class RoleAssignmentService {
+  public static allocationsUrl = '/api/role-assignment';
+
+  public judicialRoles: Role[];
+  public legalOpsRoles: Role[];
```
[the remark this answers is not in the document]
connorpgpmcelroy: Originally I did put the http calls in here but moved them to app.component.ts. After the change, this is now being used to keep the roles data. Do you want me to change this?

src/role-access/containers/allocate-role/choose-role/choose-role.component.ts
```diff
@@ -104,6 +94,10 @@ export class ChooseRoleComponent implements OnInit, OnDestroy {
     }
   }
 
+  private getOptions(roles: Role[]): OptionsModel[] {
+    return roles.map(jr => ({optionId: jr.roleId, optionValue: jr.roleName}));
```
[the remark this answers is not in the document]
connorpgpmcelroy: sorry, this is a result of the code being simplified. I will change jr to role

src/app/containers/app/app.component.ts
```diff
@@ -42,7 +45,9 @@ export class AppComponent implements OnInit, OnDestroy {
     private readonly featureService: FeatureToggleService,
     private readonly loggerService: LoggerService,
     private readonly cookieService: CookieService,
-    private readonly environmentService: EnvironmentService
+    private readonly environmentService: EnvironmentService,
+    private readonly http: HttpClient,
+    private readonly roleAssignmentService: RoleAssignmentService
```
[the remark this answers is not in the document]
rayliangatos: I just think this is not the right approach, it makes no sense to put these hardcoded role code in the App component level. As look at those services like environmentService/FeatureToggleService etc. they are some more generic component used in many places. But this roleAssignmentService seems to be used in a specific component level. Let's discuss this with Andy tomorrow.

src/app/containers/app/app.component.ts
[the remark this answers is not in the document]
rayliangatos: yes let's have a call tomorrow.

api/roleAssignment/roleAssignmentService.ts
```diff
@@ -0,0 +1,15 @@
...
+    {roleId: 'lead', roleName: 'Lead judge'},
+    {roleId: 'hearing', roleName: 'Hearing judge'}];
+  return res.send(personRoles).status(200);
+}
+
+export async function getLegalOpsRoles(req: EnhancedRequest, res: Response, next: NextFunction): Promise<any> {
+  const personRoles = [
+    {roleId: 'manager', roleName: 'Case manager'}];
+  return res.send(personRoles).status(200);
+}
```
rayliangatos, line 15:
```suggestion
export async function getTypeOfRoles(req: EnhancedRequest, res: Response, next: NextFunction): Promise<any> {
  const personRoles = [
    {roleId: 'leadJudge', roleName: 'Lead judge', roleType: 'judicial'},
    {roleId: 'hearingJudge', roleName: 'Hearing judge', roleType: 'judicial'},
    {roleId: 'caseManager', roleName: 'Case manager', roleType: 'legalOps'}];
  return res.send(personRoles).status(200);
}
```
You could merge this data in node layer, on Angular layer you don't need to call two APIs and join them, but just filter the typeOfRole by roleType.

connorpgpmcelroy: This is a good point. Was going to say I wasn't sure how the data would be returned from the API but it doesn't matter because we can merge the data within this function

PR #1303, opened 2021-09-09
fix karma coverage report

rayliangatos:
### Change description ###
a. HMCTS has set the Jenkins pipeline below by default which we can't change:
1. yarn test
2. yarn test:coverage
3. yarn test:a11y
But time is only allowed with 20mins to run the whole tests, so we have to run the test and generate the report at the same time, and we can't split the test and reporting for this time being within 20 mins. So we changed below:
1. yarn test (to run with the node test and generate the report)
2. yarn test:coverage (to run with the angula test and generate the report)
3. yarn test:a11y (no changes)

b. As the karma-parralle breaks the coverage report, so we now set the executor = 1, if executor > 1 the coverage report will be broken and it can't calculate and summarize with the right data statistics, we will check further how to ensure the karma-parralle doesn't break the coverage report.

```
[ ] Yes
[x] No
```


ernestman28 (approved):

udaydenduluri33 (approved):
@rayliangatos  thanks for fixing the issue

PR #1320, opened 2021-09-21
Global Error Handler with http interceptor

waynepHMCTS:
Jira ticket
https://tools.hmcts.net/jira/browse/EUI-3386

### Change description ###
Create Error Page for unhandled errors  by creating an interceptor that will fire up the default errorhandler, which will redirect to main error page component

```
[ ] Yes
[ x] No
```


waynepHMCTS:
The changes will stop the application for unhandled errors so all unhandled errors must be addressed

src/app/services/errorHandler/defaultErrorHandler.ts
```diff
@@ -1,13 +1,18 @@
...
+   constructor(
+     private loggerService: LoggerService,
+     private injector: Injector,
+     private zone: NgZone) { }
 
-    handleError(error: Error) {
-        this.loggerService.error(error);
+   public handleError(error: Error) {
+      this.loggerService.error(error);
+      this.router =  this.injector.get(Router);
```
udaydenduluri33, line 15: Out of curiosity Why does this need NgZone?

udaydenduluri33: Again Out of curiosity we could inject Router rather than explicitly using injector. Whats the need for it?

src/app/services/errorHandler/defaultErrorHandler.ts
```diff
@@ -1,13 +1,18 @@
...
+      this.zone.run(() => this.router.navigate(['/main-error']));
```
udaydenduluri33, line 16: Could this be moved to common constants?

src/app/services/errorHandler/defaultErrorHandler.ts
waynepHMCTS: > Again Out of curiosity we could inject Router rather than explicitly using injector. Whats the need for it?

If you inject the router is causes cyclic referencing error

waynepHMCTS: > Out of curiosity Why does this need NgZone?

NgZone has to be used because The errors in the error handlers are out of angular zone.
https://stackoverflow.com/questions/48787669/angular-navigation-in-errorhandler-only-works-with-zone-run

src/app/containers/main-error-page/main-error-page.component.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { Component, CUSTOM_ELEMENTS_SCHEMA, DebugElement, ViewChild } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { RouterTestingModule } from '@angular/router/testing';
+import { MainErrorPageComponent } from './main-error-page.component';
+
+describe('MainErrorPageComponent', () => {
+  @Component({
+    selector: `exui-app-host-dummy-component`,
```
Daniel-Lam, line 8: @waynepHMCTS - I think the backticks are meant to be single quotes instead? Not sure if it matters all that much

src/app/containers/main-error-page/main-error-page.component.html
```diff
@@ -0,0 +1,16 @@
+<exui-app-header></exui-app-header>
+<exui-page-wrapper [title]="'Error Page'">
```
Daniel-Lam, line 2: @waynepHMCTS - is it intentional to have the title showing as `'Error Page'`, with the single quote marks?

src/app/containers/main-error-page/main-error-page.component.html
```diff
@@ -0,0 +1,16 @@
...
+  <div class="govuk-width-container">
+    <main class="govuk-main-wrapper " id="main-content" role="main">
```
Daniel-Lam, line 4: @waynepHMCTS - (nit-pick): I'd take out the space after `govuk-main-wrapper`

Daniel-Lam (changes requested):
@waynepHMCTS - just a few things I've spotted, nothing major

src/app/containers/main-error-page/main-error-page.component.html
waynepHMCTS: Yes

src/app/services/errorHandler/defaultErrorHandler.ts
udaydenduluri33: `/main-error`

Daniel-Lam (dismissed):

udaydenduluri33 (approved):
@waynepHMCTS thanks for making the changes

PR #1322, opened 2021-09-22
fix EUI-3841

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-3822
https://tools.hmcts.net/jira/browse/EUI-3841

### Change description ###
case-viewer.component.ts in CCD toolkit when applyTrigger it keeps the router state of the NavigationNotifierService when webapp trigger an event from case details, it's always retrieve last obsoleted NavigationNotifierService route state data, so tidy up the data when case detail is destroyed and it can go with normal router link.

```
[ ] Yes
[x] No
```


udaydenduluri33 (dismissed):
@rayliangatos  thanks for the fix

connorpgpmcelroy (approved):
LGTM, short change

kuda1992 (dismissed):

PR #1326, opened 2021-09-24
Feature/e UI 4200 task error

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-4200?filter=-1

### Change description ###
When you claim a task if the api return an error 400 ensure `handleFatalError` function prints the error message. 

```
[ ] Yes
[x ] No
```


src/work-allocation-2/utils/work-allocation-utils.ts
```diff
@@ -59,6 +59,29 @@ export const handleFatalErrors = (status: number, navigator: Navigator, fatals?:
   }
 };
 
+export const handleTasksFatalErrors = (status: number, navigator: Navigator, fatals?: FatalRedirect[]): number => {
```
connorpgpmcelroy, line 62: Minor comment - Is this handleTaskFatalErrors necessary as handleFatalErrors is already present in the file and there is one small change for 400 errors?

connorpgpmcelroy (commented):
Unsure whether two types of error handling needed here

src/work-allocation-2/utils/work-allocation-utils.ts
kuda1992: At first glance it's not necessary however `handleFatalErrors` is a generic error function used in multiple places. Changing how we handle errors, just to cater for tasks screen will break the business rules on other screens.

udaydenduluri33: Agree with @connorpgpmcelroy  here, The generic error handling should be enough. Are you saying that this change is only for the Tasks Screen?

kuda1992: @udaydenduluri33 yes. The solution is reuse `handleFatalErrors` function and add an extra parameter to treat 400 errors from that screen differently or create a new function, I chose the later.

udaydenduluri33: Ok makes sense

udaydenduluri33 (approved):
@kuda1992  thanks for making the changes

connorpgpmcelroy (approved):

PR #1357, opened 2021-10-06
Changes to find caseworker

connorpgpmcelroy:
Mostly based on change to common library

https://tools.hmcts.net/jira/browse/EUI-4551

### Change description ###
Changing webapp to allow caseworker search and search for all people

```
[ ] Yes
[ x ] No
```


api/workAllocation2/personService.ts
```diff
@@ -21,10 +22,5 @@ export async function postFindPersonSearch(req: EnhancedRequest, res: Response,
 }
 
 export function getDefaultResult() {
```
udaydenduluri33, line 24: Can we remove this method if there is nothing we are doing here?

udaydenduluri33 (approved):
@connorpgpmcelroy  very good stuff.

kuda1992 (approved):

api/workAllocation2/personService.ts
connorpgpmcelroy: Yes, will do actually

PR #1460, opened 2021-11-11
Feature/e UI 4690 location api

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-4690

### Change description ###
Node layer to mock search locations for Hearings and Work allocation, this is a common API for both modules.
This needs to be merged in to master if work allocation needs to use it before hearing is released.

```
[ ] Yes
[x] No
```


api/locations/location.mock.ts
```diff
@@ -0,0 +1,37 @@
+/* tslint:disable:max-line-length */
+import MockAdapter from 'axios-mock-adapter';
+import {HttpMockAdapter} from '../common/httpMockAdapter';
+import {ALL_COURT_LOCATIONS} from "./data/location.mock.data";
+
+export const init = () => {
+  const mock: MockAdapter = HttpMockAdapter.getInstance();
+
```
waynepHMCTS, line 8: Is this space necessary

api/locations/models/location.model.ts
```diff
@@ -0,0 +1,26 @@
+export interface LocationModel {
```
waynepHMCTS, line 1: Is suppose to be in alphabetic order

api/locations/data/locationType.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum LocationTypeEnum {
+  // Hearing location
```
waynepHMCTS, line 2: You could name the variable with a prefix off hearingLocation unnecessary comment

api/locations/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    res.status(status).send(result);
+  } catch (error) {
+    next(error);
+  }
+}
+
+function getCourtTypeIdsByService(service: string): string {
+  const courtTypeIdsArray: [] = SERVICES_COURT_TYPE_MAPPINGS[service];
+  if (courtTypeIdsArray) {
```
waynepHMCTS, line 46: A blank line before the if statement is needed

api/locations/location.mock.ts
```diff
@@ -0,0 +1,37 @@
...
+  const getHearingsUrl = /http:\/\/rd-professional-api-aat.service.core-compute-aat.internal\/refdata\/location\/court-venues\/venue-search\?search-string=[\w]*/;
+
+  mock.onGet(getHearingsUrl).reply(config => {
+    const requestURL = config.url;
+    const searchTermStartIdx = config.url.indexOf('=');
+    const searchTermEndIdx = config.url.indexOf('&');
+    let searchTerm;
+    if (searchTermEndIdx) {
```
waynepHMCTS, line 16: Same thing

api/locations/location.mock.ts
rayliangatos: it's not necessary can remove

api/locations/models/location.model.ts
rayliangatos: aggreed but this is to match the order from the API contract, for easier maintainance if they add or remove a field then we can be easier to address where it is.

api/locations/location.mock.ts
rayliangatos: done

api/locations/index.ts
rayliangatos: not necessary but I have changed it as you like

api/locations/location.mock.ts
rayliangatos: not necessary but I have changed it as you like

api/locations/data/locationType.enum.ts
rayliangatos: I dissagreed this I would prefer not put in any necessary prefix to keep the enum key as short as possible.
For thie comment out code it's for reference because we have two location type one for Hearing and another for Work allocation.

waynepHMCTS (approved):

PR #1492, opened 2021-11-22
Bugfix/e UI 4560 case details active tab

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-4560?filter=-1

### Change description ###
save case-details, all work and my-work path name.

```
[ ] Yes
[x ] No
```


src/work-allocation-2/containers/all-work-home/all-work-home.component.spec.ts
```diff
@@ -20,12 +21,13 @@ class WrapperComponent {
   @ViewChild(AllWorkHomeComponent) public appComponentRef: AllWorkHomeComponent;
 }
 
-describe('AllWorkHomeComponent', () => {
+describe('AllWorkHomeComponent2', () => {
```
connorpgpmcelroy, line 24: Think we should change this back

connorpgpmcelroy: When we remove the release 1 code this may confuse people

src/work-allocation-2/containers/task-assignment-confirm/task-assignment-confirm.component.spec.ts
```diff
@@ -151,22 +151,22 @@ describe('TaskAssignmentConfirmComponent', () => {
   });
 
   it('should redirect to the "All work" page on cancelling task assignment', () => {
-    window.history.pushState({returnUrl: 'all-work/tasks', showAssigneeColumn: false}, '', 'all-work/tasks');
+    mockSessionStorageService.getItem.and.returnValue('all-work/tasks');
```
connorpgpmcelroy, line 154: we could write a constant for 'all-work/tasks' given that it's used a number of times

connorpgpmcelroy (approved):
Two changes I think could be made but approving as both minor points

src/work-allocation-2/containers/all-work-home/all-work-home.component.ts
```diff
@@ -40,6 +43,7 @@ export class AllWorkHomeComponent implements OnInit, OnDestroy {
 
     // Set up the page data.
     this.setupPageData(this.router.routerState.root.snapshot);
+    this.sessionStorageService.setItem(AllWorkHomeComponent.CLAIM_URL_RETURN_KEY, window.location.pathname);
```
udaydenduluri33, line 46: Why are you storing this in session Storage? Is that part of requirement?

kuda1992: @udaydenduluri33 just thought it was the optimal solution to store where the user is coming from.

udaydenduluri33: @kuda1992 I dnt think it should be stored in session storage. This should be picked from router data and sent back. Its need not be persisted as its transient information. So, I suggest please remove it. Thanks.

udaydenduluri33 (approved):

src/work-allocation-2/containers/all-work-home/all-work-home.component.ts
udaydenduluri33: @kuda thanks for making the changes

PR #1498, opened 2021-11-23
Feature/EUI 4360 book time location page

tarikdogan-hmcts:

https://tools.hmcts.net/jira/browse/EUI-4360

### Change description ###

Book your time at location page display

```
[ ] Yes
[x] No
```


tarikdogan-hmcts:
Merge into branch changed from master to  : https://github.com/hmcts/rpx-xui-webapp/tree/feature/EUI-4347-booking-ui

src/booking/containers/booking-date/booking-date.component.scss
```diff
@@ -0,0 +1,6 @@
+/deep/.date-head h1 {
```
paulhowes-HMCTS, line 1: As discussed, I don't think we should be overriding GDS guidelines and rolling our own css, however this is the first time we are using the common lib date component and without a refactor of the component itself we will not be able to add the correct styling for the date component. In future we should move toward using the gov-ui components in the common lib.

paulhowes-HMCTS (approved):
Approved with a comment regarding trying where possible to use / amend common lib govui components.

src/booking/containers/booking-date/booking-date.component.scss
Daniel-Lam: @tarikdogan-hmcts - I'm not entirely clear which `<h1>` element is being targeted. The `<xuilib-gov-uk-date>` component itself does not contain any `<h1>` tags, but it does use `<xuilib-gov-uk-fieldset>` and `<xuilib-gov-label>`, both of which contain `<h1>` tags themselves. Which of these is your target?

src/booking/containers/booking-date/booking-date.component.scss
```diff
@@ -0,0 +1,6 @@
...
+  margin-top: 0px;
+  margin-bottom: 10px;
+  font-weight: bold !important;
```
Daniel-Lam, line 4: @tarikdogan-hmcts - I would try to avoid using the `!important` override. If it's really essential, then consider doing it the GDS way by using the `govuk-!-font-weight-bold` CSS class (see https://design-system.service.gov.uk/styles/typography/). @paulhowes-HMCTS - what do you reckon?

Daniel-Lam (changes requested):
@tarikdogan-hmcts - some comments we can look at regarding the `<h1>` override.

src/booking/containers/booking-date/booking-date.component.ts
```diff
@@ -1,25 +1,84 @@
...
+  BookingTimePageText,
+} from '../../models/booking-date.enum';
+import { ErrorMessage } from '../../../app/models';
+import { GovUiConfigModel } from '@hmcts/rpx-xui-common-lib/lib/gov-ui/models';
 
 @Component({
   selector: 'exui-booking-date',
   templateUrl: './booking-date.component.html',
-  styleUrls: ['./booking-date.component.scss']
+  styleUrls: ['./booking-date.component.scss'],
```
paulhowes-HMCTS, line 19: Why the comma?

src/booking/containers/booking-date/booking-date.component.scss
```diff
@@ -0,0 +1,6 @@
+/deep/.date-head h1 {
+  margin-top: 0px;
+  margin-bottom: 10px;
+  font-weight: bold ;
```
paulhowes-HMCTS, line 4: Superfluous space between bold and ;

src/booking/models/booking-date.enum.ts
```diff
@@ -0,0 +1,13 @@
+export enum TimeOption {
+  TODAY = 'Today only (ends at midnight)',
+  WEEK = 'This week (end on Sunday at midnight )',
+  DATERANGE = 'Select a date range',
+  BOKINGSTART = 'Booking start',
+  BOKINGEND = 'Booking ends'
```
adnanakgun, line 6: Should be BOOKINGSTART and BOOKINGEND

src/booking/containers/booking-date/booking-date.component.html
```diff
@@ -1,4 +1,47 @@
...
+        {{ errorMessage.description }}
+      </div>
+      <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios">
+        <div class="govuk-radios__item" *ngFor="let item of dateInterval; let i = index">
+          <input
+            class="govuk-radios__input"
+            id="{{ 'date-' + i }}"
+            name="radioSelected"
+            type="radio"
+            [value]="item.date"
```
adnanakgun, line 22: Should the value not be an enum or some code rather than the label text?

adnanakgun (changes requested):

src/booking/containers/booking-date/booking-date.component.html
adnanakgun: Please ignore, all fine

src/booking/containers/booking-date/booking-date.component.scss
Daniel-Lam: Marking as resolved, given that this CSS rule is no longer required.

src/booking/containers/booking-date/booking-date.component.scss
Daniel-Lam: Marking as resolved, given that this CSS rule is no longer required.

src/booking/containers/booking-date/booking-date.component.ts
Daniel-Lam: Hi @paulhowes-HMCTS - resolving on your behalf since @tarikdogan-hmcts has removed `styleUrls` entirely (the SCSS file is no longer required).

src/booking/containers/booking-date/booking-date.component.scss
Daniel-Lam: Hi @paulhowes-HMCTS - resolving on your behalf since @tarikdogan-hmcts has removed the SCSS file itself (it's no longer required).

Daniel-Lam (approved):
@tarikdogan-hmcts - following our online pair programming session, I'm happy we identified the cause of the issue was an incorrect setting of the config being passed to the `xuilib-gov-uk-date` component (`isPageHeading` was set to `true` rather than `false`, which caused it to render labels between `<h1>` tags instead of the usual `<legend>` tags). This has been resolved, so the PR is approved by me.

PR #1536, opened 2021-12-03
Feature/e UI 4875 unique cases

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-4875 - Note not completely finished with regards to potential CCD fix for returning all cases

### Change description ###
Sorting by case ID
Adding header to indicate number of unique cases
Fixing pagination issue when filter is applied

```
[ ] Yes
[ x ] No
```


api/workAllocation2/util.ts
```diff
@@ -482,7 +488,7 @@ export function getSubstantiveRoles(roleAssignments: RoleAssignment[]): RoleAssi
 }
 
 // Note: array type may need to be changed depending on where pagination called
-export const paginate = (array: Case[], pageNumber: number, pageSize: number): any[] => {
+export const paginate = (array: RoleCaseData[], pageNumber: number, pageSize: number): any[] => {
```
kuda1992, line 491: Paginate is a generic function. Can we can this take an `array` generic type. Example code below

`export function paginate<T>(array: T[], pageNumber: number, pageSize: number): T[] {
  return array.slice((pageNumber - 1) * pageSize, pageNumber * pageSize);
};`

api/workAllocation2/util.ts
```diff
@@ -500,3 +506,8 @@ export async function getTypesOfWorkByUserId(path, req: express.Request): Promis
   }
   return null;
 }
+
+export function getUniqueCases(caseData: RoleCaseData[]): number {
```
kuda1992, line 510: This should be called `getUniqueCasesCount`

kuda1992 (changes requested):
Few minor changes

api/workAllocation2/util.ts
connorpgpmcelroy: good point, will do

api/workAllocation2/util.ts
connorpgpmcelroy: sure

kuda1992 (approved):
Looks good

PR #1558, opened 2021-12-09
case flag reference data

Kartheek-S:
https://tools.hmcts.net/jira/browse/EUI-4768

### Change description ###
Mock API data for the case flag reference data

```
[ ] Yes
[ x ] No
```


api/hearings/models/caseFlagReference.model.ts
```diff
@@ -0,0 +1,7 @@
+export interface CaseFlagReferenceModel {
+  id: number;
+  flag_code: string;
+  value_en: string;
+  value_cy: string;
+  category_id: number;
+}
```
rayliangatos, line 7: It's not simple as this, we need to follow the same API contract from the Hearing Flags's api LLD.
Check the response Output payload in the:
https://tools.hmcts.net/confluence/display/CF/Case+Flags+HLD+Version+1.0
This needs re-work.

api/hearings/data/caseFlagReference.mock.data.ts
```diff
@@ -0,0 +1,863 @@
+/* tslint:disable:object-literal-sort-keys */
+import { CaseFlagReferenceModel } from 'hearings/models/caseFlagReference.model';
```
rayliangatos, line 2:
```suggestion
import {CaseFlagReferenceModel} from '../models/caseFlagReference.model';
```

rayliangatos (commented):
The data hirachy looks great now, just seems something is missing, how do you call the API?

I suppose it missed the entry point in api/hearings/route, also it needs a method to call the API, if you could refer to the 

router.get('/getRefData', getRefData);

rayliangatos (commented):
good work, can get data from api locally:
http://localhost:3001/api/hearings/getCaseFlagRefData

rayliangatos (approved):

PR #1717, opened 2022-02-07
EUI-4958-welsh-in-summary

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-4958

### Change description ###
Select/unselect Welsh location and its behavior in check answer page

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/hearings/components/hearing-summary/hearing-summary.component.html
+++ b/src/hearings/components/hearing-summary/hearing-summary.component.html
@@ -1,19 +1,21 @@
 <div *ngFor="let section of template">
-  <div *ngIf="section.insetInfo" class="govuk-inset-text">
-    {{ section.insetInfo }}
-  </div>
-  <div [innerHtml]="section.sectionHTMLTitle"></div>
-  <dl class="govuk-summary-list govuk-!-margin-bottom-9">
-    <div class="govuk-summary-list__row" *ngFor="let answer of section.answers">
-      <dt class="govuk-summary-list__key">
-        {{ answer.answerTitle }}
-      </dt>
-      <dd class="govuk-summary-list__value">
-        <div [innerHtml]="answer.answerSource | transformAnswer | async"></div>
-      </dd>
-      <dd class="govuk-summary-list__actions">
-        <a *ngIf="answer.changeLink" href="javascript:void(0)" (click)="changeAnswer(answer.changeLink)">Change</a>
-      </dd>
+  <div *ngIf="section.isHiddenSource === undefined ? true : !(section.isHiddenSource | isHidden | async)">
+    <div *ngIf="section.insetInfo" class="govuk-inset-text">
+      {{ section.insetInfo }}
     </div>
-  </dl>
+    <div [innerHtml]="section.sectionHTMLTitle"></div>
+    <dl class="govuk-summary-list govuk-!-margin-bottom-9" *ngFor="let answer of section.answers">
+      <div *ngIf="answer.isHiddenSource === undefined ? true : !(answer.isHiddenSource | isHidden | async)" class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          {{ answer.answerTitle }}
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <div [innerHtml]="answer.answerSource | transformAnswer | async"></div>
+        </dd>
+        <dd class="govuk-summary-list__actions">
+          <a *ngIf="answer.changeLink" href="javascript:void(0)" (click)="changeAnswer(answer.changeLink)">Change</a>
+        </dd>
+      </div>
+    </dl>
+  </div>
 </div>
--- a/src/hearings/pipes/hearings.pipes.module.ts
+++ b/src/hearings/pipes/hearings.pipes.module.ts
@@ -1,16 +1,19 @@
-import { CUSTOM_ELEMENTS_SCHEMA, NgModule } from '@angular/core';
+import {CUSTOM_ELEMENTS_SCHEMA, NgModule} from '@angular/core';
 import {HearingAnswersPipe} from './hearing-answers.pipe';
-import { HearingsBadgePipe } from './hearings-badge.pipe';
+import {HearingsBadgePipe} from './hearings-badge.pipe';
+import {ShowHidePipe} from './show-hide.pipe';
 
 @NgModule({
   declarations: [
     HearingsBadgePipe,
-    HearingAnswersPipe
+    HearingAnswersPipe,
+    ShowHidePipe
   ],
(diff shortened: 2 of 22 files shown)


src/hearings/pipes/hearings.pipes.module.ts
```diff
@@ -1,16 +1,19 @@
...
+import {ShowHidePipe} from './show-hide.pipe';
 
 @NgModule({
   declarations: [
     HearingsBadgePipe,
-    HearingAnswersPipe
+    HearingAnswersPipe,
+    ShowHidePipe
   ],
   schemas: [CUSTOM_ELEMENTS_SCHEMA],
```
kuda1992, line 12: Do you need `CUSTOM_ELEMENTS_SCHEMA` in a `pipe.module`?

src/hearings/pipes/show-hide.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+import {WelshHiddenConverter} from '../converters/welsh.hidden.converter';
+import {IsHiddenSource} from '../models/hearings.enum';
+import * as fromHearingStore from '../store';
+
+@Pipe({
+  name: 'isHidden'
+})
+export class ShowHidePipe implements PipeTransform {
+
+  constructor(protected readonly hearingStore: Store<fromHearingStore.State>) {
```
kuda1992, line 15: Do we need to inject the store into a pipe?. Pipe are meant for simple data transformations and pure.

src/hearings/pipes/show-hide.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public transform(isHiddenSource: IsHiddenSource): Observable<boolean> {
+    let converter: AbstractHiddenConverter = new DefaultHiddenConverter(this.hearingStore);
+    switch (isHiddenSource) {
```
kuda1992, line 20: And do we need a switch here?

kuda1992 (changes requested):
Just a few points to look at.

src/hearings/components/hearing-summary/hearing-summary.component.html
```diff
@@ -1,19 +1,24 @@
...
-        <a *ngIf="answer.changeLink" href="javascript:void(0)" (click)="changeAnswer(answer.changeLink)">Change</a>
-      </dd>
+  <div *ngIf="section.isHiddenSource === undefined ? true : !(section.isHiddenSource | isHidden | async)">
+    <div *ngIf="section.insetInfo" class="govuk-inset-text">
+      {{ section.insetInfo }}
+    </div>
+    <div [innerHtml]="section.sectionHTMLTitle"></div>
+    <div class="govuk-summary-list govuk-!-margin-bottom-9">
+      <ng-container *ngFor="let answer of section.answers">
+        <div *ngIf="answer.isHiddenSource === undefined ? true : !(answer.isHiddenSource | isHidden | async)"
```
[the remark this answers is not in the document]
rayliangatos: This is an async pipe not observables

src/hearings/pipes/hearings.pipes.module.ts
rayliangatos: It's not needed

src/hearings/pipes/show-hide.pipe.ts
rayliangatos: I think that's fine we are using pipe to subscribe data in the store, pipe can also do complex things it's not exactly for simple data transformation.

src/hearings/pipes/show-hide.pipe.ts
rayliangatos: We will later have more hidden logic here on view edit mode, so it's not a single case.

phillip-whitaker-hmcts (approved):
LGTM+

src/hearings/pipes/show-hide.pipe.ts
kuda1992: Okay sounds reasonable to me

src/hearings/components/hearing-summary/hearing-summary.component.html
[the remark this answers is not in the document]
kuda1992: Okay

src/hearings/pipes/show-hide.pipe.ts
kuda1992: Yes pipes can do complicated things but in general should be simple and pure. Given the same input they will should return the same result. We should at least pass the state of the store into the `transform()` method not inject the whole store.

rayliangatos: I can't get your point or I don't agree that pass the state into the transform() method, it will be then very heavy on the html and its specific ts. It's not a problem to inject the store into the pipe, coz that's how that store is injected or how it is used, it could be unsubscribed it from the pipe on OnDestroy if any subscription is made, but giving that we are using aync pipe it has no side effect. Also if you inject the store that is just an reference to the store, it's not actually load all the data on the pipe.

kuda1992: This is complicated for no reason. Your pipe responsibility is to hide or show, but your pipe has side effects and contains state. Pipes are transformational functions. In your case the pipe transform should take hearing state and return true or false.  E.g. html like this 
`<div *ngIf="section.isHiddenSource | isHidden:  hearingState"></div>`

ts 
`public transform(isHiddenSource: IsHiddenSource, state: HearingState): boolean {
    let converter: AbstractHiddenConverter = new DefaultHiddenConverter(state);
    switch (isHiddenSource) {
      case IsHiddenSource.WELSH_LOCATION:
        converter = new WelshHiddenConverter(state);
        break;
      default:
        break;
    }
    return converter.transformHidden();
  }`

rayliangatos: This is not completed at all, store from the pipe here is an immutable single source of data, I can't get that why you said it has side effect, we are not dispatching data in the pipe, we are just using async pipe to show the data.
In you sample, it pass the state from the html that makes no benefit at all, it increases the complexity in the html and its type script. and compare the lines of code it seems more complex.
This approach has been reviewed and approved by Andy.
Unless you could specify what the 'side effect' is we will stick with this solution.

rayliangatos: here is an example for inject store in pipe:
https://stackoverflow.com/questions/59130911/is-it-recommended-to-subscribe-to-the-store-ngrx-inside-a-pipe

kuda1992 (approved):
Looks all good to me @rayliangatos . Thanks for making the changs 😊

PR #1753, opened 2022-02-16
add a get hearing actuals endpoint

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-5242?filter=-1

### Change description ###
Created a new endpoint for hearing actuals

```
[ ] Yes
[ x] No
```

### DIFF
--- a/api/hearings/data/hearing-actuals.mock.data.ts
+++ b/api/hearings/data/hearing-actuals.mock.data.ts
@@ -1,0 +1,123 @@
+import { Party } from '../../../src/hearings/models/hearingActuals.model';
+import {
+  ActualDayParty,
+  ActualHearingDay,
+  HearingActualsModel,
+  HearingResult,
+  PlannedHearing
+} from '../models/hearingActuals.model';
+import { HMCStatus } from '../models/hearings.enum';
+import { getRandomElement, getRandomProperty } from './mock-data-utils';
+
+export function generateHearingActuals(numberOfActuals: number): HearingActualsModel[] {
+
+  const hearingActuals: HearingActualsModel[] = [];
+
+  for (let i = 0; i < numberOfActuals; i++) {
+    hearingActuals.push({
+      hearingActuals: {
+        hearingOutcome: {
+          hearingFinalFlag: getRandomElement([true, false]),
+          hearingResult: getRandomProperty(HearingResult) as unknown as HearingResult,
+          hearingResultDate: getRandomElement(['2019-01-01', '2019-01-02', '2019-01-03']),
+          hearingResultReasonType: '',
+          hearingType: getRandomElement([
+            'Pre-hearing review',
+            'Case management preliminary hearing - open',
+            'Full hearing',
+            'Initial hearing',
+            'Directions hearing']
+          ),
+        },
+        actualHearingDays: generateActualHearingDay(2),
+      },
+      hearingPlanned: {
+        plannedHearingType: getRandomElement([1, 2, 3, 4, 5]),
+        plannedHearingDays: generatePlannedHearings(4),
+      },
+      hmcStatus: getRandomProperty(HMCStatus) as unknown as HMCStatus,
+    });
+  }
+  return hearingActuals;
+}
+
+function generateParties(numberOfParties: number): Party[] {
+  const parties: Party[] = [];
+
+  for (let i = 0; i < numberOfParties; i++) {
+    const id = getRandomElement([1, 2, 3, 4, 5]);
+    parties.push({
+      individualDetails: {
+        title: getRandomElement(['Mr', 'Mrs', 'Miss', 'Ms']),
+        firstName: getRandomElement(['John', 'Jane', 'Mary', 'Bob', 'Tom']),
+        lastName: getRandomElement(['Smith', 'Jones', 'Brown', 'Smith', 'Jones']),
+      },
+      organisationDetails: {
+        cftOrganisationID: getRandomElement(['12345', '54321', '67890']),
+        name: getRandomElement(['Company A', 'Company B', 'Company C', 'Company D', 'Company E']),
+      },
+      partyId: id,
+      partyRole: id,
+      partyChannelSubType: getRandomElement(['Email', 'Phone', 'Fax', 'Letter', 'Other']),
+    });
+  }
+  return parties;
+}
+
+function generateActualDayParties(numberOfDayParties: number): ActualDayParty[] {
+  const dayParties: ActualDayParty[] = [];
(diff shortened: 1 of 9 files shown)


api/hearings/data/hearing-actuals.mock.data.ts
```diff
@@ -0,0 +1,123 @@
+import { Party } from '../../../src/hearings/models/hearingActuals.model';
```
rayliangatos, line 1: api refer to src model

api/hearings/data/hearing-actuals.mock.data.ts
```diff
@@ -0,0 +1,123 @@
...
+    const plannedHearing: PlannedHearing = {
+      plannedStartTime: '2021-03-12T09:00:00.000+0000',
+      plannedEndTime: '2021-03-13T10:00:00.000+0000',
+      parties: generateParties(2),
+    };
+    plannedHearings.push(plannedHearing);
+  }
+
+  return plannedHearings;
+}
```
rayliangatos, line 123: Apprieciated for generating random value for a key in the models, but this has been proved in other projects like NoC/share case etc., it doesn't quite work well for demoing LIVE alike data, it is hard to maintain on development, not actually use the benefits for typescript models check, for those timings in requirement it can't be past time but here is random time including past, and also it can't reuse those reference data like hearing type, hearing result which they are from enum types etc. I would prefer this is just a hard coded mock data with using typescript type check on the models, just like we have done for get hearing details.

api/hearings/hearing.mock.ts
```diff
@@ -2,16 +2,20 @@ import { AxiosRequestConfig } from 'axios';
...
 import { ALL_REF_DATA } from './data/reference.mock.data';
 import { SERVICE_HEARING_VALUES } from './data/serviceHearingValues.mock.data';
-import { HEARING_RESPONSE_RESULT } from './data/hearingResponse.mock.data';
 
 export const init = () => {
   const mock: MockAdapter = HttpMockAdapter.getInstance();
 
   const getHearingsUrl = /https:\/\/hearings.aat.service.core-compute-aat.internal\/hearings\/[0-9]{16}/;
 
+  const getHearingsActualsUrl = /https:\/\/hearings.aat.service.core-compute-aat.internal\/actuals\/[0-9]{16}/;
```
rayliangatos, line 17: hearing id is not 16 digits, and we don't know what it looks like yet, so could we use [\w]* as the getHearingInfoUrl below?

api/hearings/models/hearingActuals.model.ts
```diff
@@ -0,0 +1,89 @@
...
+  hearingEndTime: string;
+  pauseDateTimes: PauseDateTime[];
+  actualDayParties: ActualDayParty[];
+}
+
+export enum HearingResult {
+  COMPLETED = 'COMPLETED',
+  ADJOURNED = 'ADJOURNED',
+  CANCELLED = 'CANCELLED',
+}
```
rayliangatos, line 64: could we move all those enums to hearings.enum.ts? The benefit is when we create enum, we need to check whether it's existing.

api/hearings/models/hearingActuals.model.ts
```diff
@@ -0,0 +1,89 @@
...
+interface IndividualDetail {
+  title: string;
+  firstName: string;
+  lastName: string;
+}
+
+interface OrganisationDetails {
+  name: string;
+  cftOrganisationID: string;
+}
```
rayliangatos, line 37: These doesn't have export keyword are they privately use in this model only?

api/hearings/models/hearingActuals.model.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export interface HearingActuals {
+  hearingOutcome: HearingOutcome;
+  actualHearingDays: ActualHearingDay[];
+}
+
+interface HearingOutcome {
```
rayliangatos, line 44: no export?

src/hearings/models/hearingActuals.model.ts
```diff
@@ -0,0 +1,89 @@
...
+interface ActualDayParty {
+  actualPartyId: number;
+  didNotAttendFlag: boolean;
+  actualIndividualDetails: ActualIndividualDetail;
+  actualOrganisationDetails: ActualOrganisationDetails;
+  partyChannelSubType: string;
+  partyId: number;
+  partyRole: number;
+  representedParty: number;
+}
```
rayliangatos, line 89: same comments as the one under api

api/hearings/data/hearing-actuals.mock.data.ts
kuda1992: Hmm I am not sure how it's hard to maintain. Probably the easiest way to create alot of fake data. About using typescript model checks I am not sure what you mean there. I am using the actual enums to generate random values. And the point about times and dates you could always create a helper function which creates times within a specific range.

rayliangatos: We used the existing faker plugin before to generate random value in mock, that proved it doesn't quite work and doesn't match the prototype we are provided, then we removed those 'Random' value logic, what we are trying to achieve here is simple, just create the hard code copy data from the prototype, from here https://ucd-xui2-prototype.herokuapp.com/hearing-actuals/version3/actuals-categories. In one of this scenario you have to control those start time < pause time < resume time < end time, and there are much more reference data you could see coming from the prototyp. It's not necessary to spend so much effort on this 'Random' value in mock data.

api/hearings/models/hearingActuals.model.ts
```diff
@@ -0,0 +1,83 @@
...
+export interface ActualDayParty {
+  actualPartyId: number;
+  didNotAttendFlag: boolean;
+  actualIndividualDetails: ActualIndividualDetail;
+  actualOrganisationDetails: ActualOrganisationDetails;
+  partyChannelSubType: string;
+  partyId: number;
+  partyRole: number;
+  representedParty: number;
+}
```
rayliangatos, line 83: It's okay to put all those models together, could we name it with suffix xxxModel to consist with naming convention for all these models, just like the HearingActualsModel you have done.

api/hearings/data/hearing-actuals.mock.data.ts
kuda1992: Sure I will change it. I have all these helper functions already fyi about the comment  on effort.

rayliangatos (approved):
thanks addressing those issue, looks good to me.

Kartheek-S (approved):

PR #1771, opened 2022-02-21
EUI-5246 Create mock API for post hearing actuals

johnbenjamin-hmcts:
https://tools.hmcts.net/jira/browse/EUI-5246

### Change description ###
Create mock API for post hearing actuals

```
[ ] Yes
[X] No
```

### DIFF
--- a/api/hearings/index.ts
+++ b/api/hearings/index.ts
@@ -135,8 +135,8 @@ export async function submitHearingActuals(req: EnhancedRequest, res: Response, 
   const hearingId = req.query.hearingId;
   const markupPath = `${hearingsUrl}/hearingActualsCompletion/${hearingId}`;
   try {
-    const { status, data }: { status: number, data: any } = await handlePost(markupPath, reqBody, req);
-    res.status(status).send(data);
+    const { status }: { status: number } = await handlePost(markupPath, reqBody, req);
+    res.sendStatus(status);
   } catch (error) {
     next(error);
   }
--- a/api/hearings/index.ts
+++ b/api/hearings/index.ts
@@ -126,3 +126,18 @@ export async function updateHearingActuals(req: EnhancedRequest, res: Response, 
     next(error);
   }
 }
+
+/**
+ * submitHearingActuals - submit hearing actuals
+ */
+export async function submitHearingActuals(req: EnhancedRequest, res: Response, next: NextFunction) {
+  const reqBody = req.body;
+  const hearingId = req.query.hearingId;
+  const markupPath = `${hearingsUrl}/hearingActualsCompletion/${hearingId}`;
+  try {
+    const { status, data }: { status: number, data: any } = await handlePost(markupPath, reqBody, req);
+    res.status(status).send(data);
+  } catch (error) {
+    next(error);
+  }
+}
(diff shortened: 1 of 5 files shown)


src/hearings/services/hearings.service.ts
```diff
@@ -52,4 +52,8 @@ export class HearingsService {
   public updateHearingActuals(hearingId: string, hearingActualsModel: HearingActualsModel): Observable<HearingActualsModel> {
     return this.http.put<HearingActualsModel>(`api/hearings/hearingActuals/${hearingId}`, hearingActualsModel);
   }
+
+  public submitHearingActuals(hearingId: string): Observable<any> {
```
rayliangatos, line 56: The any needs to be replaced with the actual model.

api/hearings/index.ts
```diff
@@ -126,3 +126,18 @@ export async function updateHearingActuals(req: EnhancedRequest, res: Response,
...
+
+/**
+ * submitHearingActuals - submit hearing actuals
+ */
+export async function submitHearingActuals(req: EnhancedRequest, res: Response, next: NextFunction) {
+  const reqBody = req.body;
+  const hearingId = req.query.hearingId;
+  const markupPath = `${hearingsUrl}/hearingActualsCompletion/${hearingId}`;
+  try {
+    const { status, data }: { status: number, data: any } = await handlePost(markupPath, reqBody, req);
```
rayliangatos, line 138: data: any if that can be a model? could you check the LLD API contract?

src/hearings/services/hearings.service.ts
```diff
@@ -52,4 +52,8 @@ export class HearingsService {
...
+    return this.http.post<any>(`api/hearings/hearingActualsCompletion/${hearingId}`, {});
```
rayliangatos, line 57: as chat body will be null as the Back end api do not accept any body including {}.

api/hearings/index.ts
```diff
@@ -126,3 +126,18 @@ export async function updateHearingActuals(req: EnhancedRequest, res: Response,
     next(error);
   }
 }
+
+/**
+ * submitHearingActuals - submit hearing actuals
+ */
+export async function submitHearingActuals(req: EnhancedRequest, res: Response, next: NextFunction) {
+  const reqBody = req.body;
```
rayliangatos, line 134: reqBody couldn't set as null as per API contract.

src/hearings/services/hearings.service.ts
johnbenjamin-hmcts: @rayliangatos - as discussed I have updated the code to use null.

api/hearings/index.ts
johnbenjamin-hmcts: @rayliangatos - as discussed I have updated the code to use null.

src/hearings/services/hearings.service.ts
johnbenjamin-hmcts: I have updated the code to use number instead of any, as the API returns only the status.

johnbenjamin-hmcts: I have updated the code to use number instead of any, as the API returns only the status.

rayliangatos (approved):

Kartheek-S (approved):

PR #1784, opened 2022-02-23
Feature/e UI 4933 amend journey generic

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-4933
https://tools.hmcts.net/jira/browse/EUI-5291

### Change description ###
Amend hearing request generic implementation

```
[ ] Yes
[x] No
```


src/hearings/store/effects/hearing-request.effects.ts
```diff
@@ -61,32 +62,52 @@ export class HearingRequestEffects {
   public continueNavigation$ = this.actions$.pipe(
     ofType(hearingRequestActions.UPDATE_HEARING_REQUEST),
     tap(() => {
-      const nextPage = this.pageFlow.getNextPage(this.screenNavigations$);
+      let nextPage;
```
johnbenjamin-hmcts, line 65: The existing code should be fine? I think nextPage can remain as a const and its value can be fetched just above switch statement.

johnbenjamin-hmcts (approved):
Approved

src/hearings/store/effects/hearing-request.effects.ts
rayliangatos: the reason to change this to let is because some of the logic is changed for amend journey, so on VIEW_EDIT mode I don't want the this.pageFlow.getNextPage(this.screenNavigations$); to be called at first as it's not needed for amend journey. we can use the preview code but the problem is this is redudant call for VIEW_EDIT mode which I want to avoid, hope that make sense.

src/hearings/store/effects/hearing-request.effects.ts
```diff
@@ -61,32 +62,52 @@ export class HearingRequestEffects {
...
         case Mode.CREATE:
+          nextPage = this.pageFlow.getNextPage(this.screenNavigations$);
           if (nextPage) {
             return this.router.navigate(['hearings', 'request', nextPage]);
           } else {
             throw new Error('Next page not found');
           }
         case Mode.CREATE_EDIT:
+          nextPage = this.pageFlow.getNextPage(this.screenNavigations$);
           if (nextPage === 'hearing-welsh') {
```
SivaprakasamSivaprakasam, line 76: can we make use of enum instead of hard-coded string

SivaprakasamSivaprakasam (approved):
approved minor comments.

src/hearings/store/effects/hearing-request.effects.ts
rayliangatos: thanks this is good point I will make a change

PR #1841, opened 2022-03-07
Bug/e UI 5448 pact

udaydenduluri33:

https://tools.hmcts.net/jira/browse/EUI-5448

### Change description ###

Pass the caseType to the payload while creating the Role Assignments including Exclusions

```
[ ] Yes
[ X ] No
```


connorpgpmcelroy (approved):
LGTM

api/roleAccess/exclusionService.ts
```diff
@@ -71,6 +70,7 @@ export function prepareExclusionBody(currentUserId: string, assigneeId: string,
         caseId: body.caseId,
         jurisdiction: body.jurisdiction,
         notes: body.exclusionDescription,
+        caseType: body.caseType,
```
scott-turner-hmcts, line 73: Destructuring and property value shorthand could be used here to avoid repetition and simplify code e.g:

`const { caseId, caseType, exclusionDescription, jurisdiction } = body;`

...

```
{
  caseId,
  jurisdiction,
  notes: exclusionDescription,
  caseType,
}
```

udaydenduluri33: Yes but it would end up something like this
`      attributes: {
        caseId,
        jurisdiction,
        notes: body.exclusionDescription,
        caseType,
      },`

udaydenduluri33: Agree that the idea is right if all the property names match

src/role-access/containers/allocate-role/allocate-role-home/allocate-role-home.component.ts
```diff
@@ -95,6 +96,7 @@ export class AllocateRoleHomeComponent implements OnInit, OnDestroy {
     if (this.route.snapshot.queryParams) {
       this.caseId = this.route.snapshot.queryParams.caseId ? this.route.snapshot.queryParams.caseId : null;
       this.jurisdiction = this.route.snapshot.queryParams.jurisdiction ? this.route.snapshot.queryParams.jurisdiction : null;
+      this.caseType = this.route.snapshot.queryParams.caseType ? this.route.snapshot.queryParams.caseType : null;
```
scott-turner-hmcts, line 99: there are lots of repeated lines in this block`this.route.snapshot.queryParams` destructuring could be used to avoid this and improve readability e.g:

`const { caseId, caseType, ... } = this.route.snapshot.queryParams`

`this.caseType = caseType ? caseType : null;`

api/roleAccess/models/allocate-role-state-data.interface.ts
```diff
@@ -2,6 +2,7 @@ import { Person } from '@hmcts/rpx-xui-common-lib/lib/models/person.model';
 import { AllocateTo, DurationOfRole, Period, RoleCategory, SpecificRole} from './allocate-role.enum';
 
 export interface AllocateRoleData {
+  caseType: string;
```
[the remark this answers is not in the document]
udaydenduluri33: Done

src/role-access/containers/allocate-role/allocate-role-home/allocate-role-home.component.ts
udaydenduluri33: Agree, good point will do it

scott-turner-hmcts (approved):
thanks for changes looks good

PR #1921, opened 2022-03-30
feature/EUI-4447

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-4447

### Change description ###
DWP user hearings tab

```
[ ] Yes
[x ] No
```


johnbenjamin-hmcts (approved):
Approved

src/cases/containers/case-hearings/case-hearings.component.ts
```diff
@@ -73,15 +88,24 @@ export class CaseHearingsComponent implements OnInit, OnDestroy {
     if (this.hearingsActions.includes(Actions.CREATE)) {
       this.hasRequestAction = true;
     }
+    this.isOgdRole$ = this.roleCategoryMappingService.getUserRoleCategory(this.userRoles$).pipe(map(userRole => userRole === UserRole.Ogd));
```
SivaprakasamSivaprakasam, line 91: Is that fetch required all the time regardless of whether the user has ogd role or not?

SivaprakasamSivaprakasam (approved):
@rayliangatos just a minor comment

src/cases/containers/case-hearings/case-hearings.component.ts
rayliangatos: it will be checked all the time from async pipe to see if the user has ogd role or not, ogd will be ahead of case worker role /judicial role check, if user has ogd role like dwp, he will be showed with the limited ogd view, otherwise it will be following existing role permission rule.

PR #1959, opened 2022-04-20
Feature/e UI 5679

rayliangatos:
https://tools.hmcts.net/jira/browse/EUI-5679

### Change description ###
Integrate with LoV APIs

```
[ ] Yes
[x] No
```


johnbenjamin-hmcts (approved):
Approved

Kartheek-S (approved):

src/hearings/containers/request-hearing/hearing-panel/hearing-panel.component.ts
```diff
@@ -201,10 +200,13 @@ export class HearingPanelComponent extends RequestHearingPageFlow implements OnI
         key: control.value.key,
         value_en: control.value.value_en,
         value_cy: control.value.value_cy,
-        hintText_EN: control.value.hintText_EN,
-        hintTextCY: control.value.hintTextCY,
-        order: control.value.order,
-        parentKey: control.value.parentKey,
+        hint_text_en: control.value.hint_text_en,
```
SivaprakasamSivaprakasam, line 203: is it worth trying to use {...control.value} instead of each key assignment as the source and destination object holding the same keys

src/hearings/containers/request-hearing/hearing-change-reason/hearing-change-reason.component.ts
```diff
@@ -99,12 +100,13 @@ export class HearingChangeReasonComponent extends RequestHearingPageFlow impleme
           key: reason.value.key,
           value_en: reason.value.value_en,
           value_cy: reason.value.value_cy,
-          hintText_EN: reason.value.hintText_EN,
-          hintTextCY: reason.value.hintTextCY,
-          order: reason.value.order,
-          parentKey: reason.value.parentKey,
+          hint_text_en: reason.value.hint_text_en,
```
SivaprakasamSivaprakasam, line 103: is it worth trying to use {...control.value} instead of each key assignment as the source and destination object holding the same keys

SivaprakasamSivaprakasam (approved):
approved with minor comment

src/hearings/containers/request-hearing/hearing-panel/hearing-panel.component.ts
rayliangatos: thanks that is a good point, absolutely we can do that, however I will keep the existing code for data transformation to avoid retest the whole journey.

src/hearings/containers/request-hearing/hearing-change-reason/hearing-change-reason.component.ts
rayliangatos: thanks that is a good point, absolutely we can do that, however I will keep the existing code for data transformation to avoid retest the whole journey.

PR #2006, opened 2022-05-16
EUI-5859-put-actual

rayliangatos:
### Change description ###

```
[ ] Yes
[ ] No
```


Kartheek-S (approved):

johnbenjamin-hmcts (approved):

src/hearings/models/hearingActualsMainModel.ts
```diff
@@ -27,14 +27,13 @@ export interface OrganisationDetailsModel {
...
+  partyID: string;
   partyRole: string;
   individualDetails?: IndividualDetailsModel;
   organisationDetails?: OrganisationDetailsModel;
   partyChannelSubType: string;
 }
 
-
 export interface ActualDayPartyModel {
   actualPartyId: string;
```
SivaprakasamSivaprakasam, line 38: it's more of just checking that we have just changed from partyId to PartyID for planningDayParty model but not for ActualDayParty modal, so will this be handled later on ?

SivaprakasamSivaprakasam (commented):
just a minor comment

SivaprakasamSivaprakasam (approved):

src/hearings/models/hearingActualsMainModel.ts
rayliangatos: this is to map with real API, in back end they use actualPartyId, but another one they use PartyID, which is insistent but we have to map with them.

PR #2025, opened 2022-05-24
Pact contract update inline JRD api change, authorisation service cod…

sreekanthpuligadda:
https://tools.hmcts.net/jira/browse/EUI-5918

### Change description ###

Pact contract update inline with JRD api change https://tools.hmcts.net/jira/browse/RDCC-4447

```
[ ] Yes
[ x] No
```


udaydenduluri33 (dismissed):
@sreekanthpuligadda  thanks for fixing it

PR #2063, opened 2022-06-08
EUI-5981: Added css class for particular row

Surekha-HMCTS:
### Change description ###

EUI-5981: Added css class for particular row


src/work-allocation-2/components/work-field/work-field.component.ts
```diff
@@ -39,4 +39,11 @@ export class WorkFieldComponent {
     return null;
   }
 
+  public getCss(workField): string {
```
udaydenduluri33, line 42: We dont need a function. Please use ngClass and template base

tarikdogan-hmcts (approved):

PR #2242, opened 2022-08-11
Feature/EUI 5672 welsh noc

SivaprakasamSivaprakasam:

https://tools.hmcts.net/jira/browse/EUI-5672

### Change description ###

Welsh language - Notice of change

```
[ ] Yes
[ x] No
```


src/noc/containers/noc-affirmation/noc-affirmation.component.html
```diff
@@ -2,16 +2,16 @@
      [ngClass]="{'govuk-form-group--error': validationErrors$ | async}">
   <div class="govuk-form-group">
     <fieldset class="govuk-fieldset" aria-describedby="sign-affirmation">
-      <p>You should tick to 'sign' when:</p>
+      <p>{{"You should tick to 'sign' when:" | rpxTranslate}}</p>
       <ul class="govuk-list govuk-list--bullet">
-        <li>You're satisfied that all these details are accurate and match what is written on the case</li>
-        <li>You have served notice of this change on every party to the case, including the former legal
-          representative (if there was one)
+        <li>{{'Youre satisfied that all these details are accurate and match what is written on the case' | rpxTranslate}}</li>
```
MateiHMCTS, line 7: Should be 
{{'You\'re satisfied that all these details are accurate and match what is written on the case' | rpxTranslate}}

SivaprakasamSivaprakasam: as it contains the single quote inside the text so it will be the problem around the text trans and that's why have to use different format if it contains '

src/noc/noc.module.ts
```diff
@@ -37,7 +38,8 @@ import { effects, reducers } from './store';
       FormsModule,
       ReactiveFormsModule,
       UtilsModule,
-      ExuiCommonLibModule
+      ExuiCommonLibModule,
+      RpxTranslationModule.forChild(),
```
MateiHMCTS, line 42: not necessary; on the base branch the SharedModule already imports and exports RpxTranslationModule

SivaprakasamSivaprakasam: i tried without adding but it doesn't recognize the rpxTranslate so I have to add this.

src/noc/containers/noc-affirmation/noc-affirmation.component.html
MateiHMCTS: I am saying that Youre should be You're

src/noc/noc.module.ts
MateiHMCTS: should work without.
If you have the SharedModule imported, you don't need to import RpxTranslationModule as you already import it (SharedModule exports RpxTranslationModule)

MateiHMCTS (approved):

Kartheek-S (approved):

PR #2426, opened 2022-10-24
Bug/EUI 6931

DavidJayakumar:

https://tools.hmcts.net/jira/browse/EUI-6931

### Change description ###

Offset timezone in my access page

```
[ ] Yes
[x] No
```


src/role-access/services/allocate-role.service.ts
```diff
@@ -25,7 +26,11 @@ export class AllocateRoleService {
     }
   }
 
-  public specificAccessApproval(specificAccessStateData: SpecificAccessStateData, period: Period): Observable<any> {
+  public specificAccessApproval(specificAccessStateData: SpecificAccessStateData, dtperiod: Period): Observable<any> {
+    const period = {
+      startDate : this.durationService.setUTCTimezone(dtperiod.startDate),
```
udaydenduluri33, line 31: There can be case where startDate / endDate is not available. Do you need to check for it?

Kartheek-S (approved):

udaydenduluri33 (approved):

PR #2427, opened 2022-10-24
bug/EUI-6975 fix

Kartheek-S:
https://tools.hmcts.net/jira/browse/EUI-6975

### Change description ###
User Excluded from case accessing same case via 16 digits reference within same session

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/cases/components/roles-and-access/roles-and-access.component.ts
+++ b/src/cases/components/roles-and-access/roles-and-access.component.ts
@@ -60,7 +60,7 @@ export class RolesAndAccessComponent implements OnInit, OnChanges {
   }
 
   public removeCashedCase(): void {
-    this.caseNotifier.cachedCaseView = null;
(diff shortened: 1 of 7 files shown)


src/cases/components/roles-and-access/roles-and-access.component.ts
```diff
@@ -55,6 +59,10 @@ export class RolesAndAccessComponent implements OnInit, OnChanges {
     }
   }
 
+  public removeCashedCase(): void {
+    this.caseNotifier.cachedCaseView = null;
```
udaydenduluri33, line 63: Minor change - please introduce a function removeCachedCase in CaseNotifier instead if marking it to null

udaydenduluri33 (approved):

DavidJayakumar (approved):

PR #2468, opened 2022-11-08
EUI-7088 Hearings Private Law Integration V2

johnbenjamin-hmcts:
https://tools.hmcts.net/jira/browse/EUI-7088

### Change description ###
Refactor Hearings Config to allow Private Law to have the API URL that they need to use

```
[ ] Yes
[X] No
```

### DIFF
--- a/config/custom-environment-variables.json
+++ b/config/custom-environment-variables.json
@@ -40,10 +40,12 @@
     "hearings": {
 			"hearingsJurisdictions": "SSCS,PRIVATELAW",
 			"sscs": {
-				"serviceApi": "SERVICES_SSCS_HEARINGS_COMPONENT_API"
+				"serviceApi": "SERVICE_HEARINGS_COMPONENT_API.sscs",
+				"caseTypes": "Appeal"
 			},
-			"privateLaw": {
-				"serviceApi": "SERVICES_PRIVATELAW_HEARINGS_COMPONENT_API"
+			"privatelaw": {
+				"serviceApi": "SERVICE_HEARINGS_COMPONENT_API.privatelaw",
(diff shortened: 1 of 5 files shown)


config/custom-environment-variables.json
```diff
@@ -40,10 +40,12 @@
     "hearings": {
 			"hearingsJurisdictions": "SSCS,PRIVATELAW",
 			"sscs": {
-				"serviceApi": "SERVICES_SSCS_HEARINGS_COMPONENT_API"
+				"serviceApi": "SERVICE_HEARINGS_COMPONENT_API.sscs",
+				"caseTypes": "Appeal"
```
andywilkinshmcts, line 44: The SSCS case type should be "Benefit",
 https://github.com/hmcts/sscs-ccd-definitions/blob/master/benefit/data/sheets/CaseType/CaseType.json

The private law case type(s) need to be confirmed with private law, but appears to be 
PRLAPPS
https://github.com/hmcts/prl-ccd-definitions/blob/master/definitions/private-law/json/CaseType.json

andywilkinshmcts (changes requested):
Might as well get the case types correct

config/custom-environment-variables.json
johnbenjamin-hmcts: Thank you @andywilkinshmcts, I checked the response of the API call /data/internal/cases/[CASE_ID] on the DEMO environment for both SSCS and PrivateLaw and can confirm that the case types you mentioned are correct.

I have updated the code accordingly.

andywilkinshmcts (approved):

tarikdogan-hmcts (approved):

PR #2509, opened 2022-11-23
EUI-7224 Menu items fix

johnbenjamin-hmcts:
https://tools.hmcts.net/jira/browse/EUI-7224

### Change description ###
WA R3 feature toggled off, missing menu item Find case

```
[ ] Yes
[X] No
```


udaydenduluri33 (commented):
I can see that you are basing your logic to see if Global Search is enabled
How can you verify if the particular service is enabled Global search?

Mo-Lala84 (approved):
approved

johnbenjamin-hmcts:
> 

We are displaying 16 digit search box only if the global search is enabled irrespective of the service.
If the global search is disabled then we show find case right aligned and do not show 16 digit search box.

udaydenduluri33 (approved):

PR #2651, opened 2023-01-09
Set the first service as default value

DavidJayakumar:

https://tools.hmcts.net/jira/browse/EUI-7352

### Change description ###

Service - default to first service in drop down list for user

```
[ ] Yes
[x] No
```


src/work-allocation/components/task-manager-filter/task-manager-filter.component.ts
```diff
@@ -223,6 +219,10 @@ export class TaskManagerFilterComponent implements OnInit, OnDestroy {
           name: 'role',
           value: [this.roleType]
         },
+        {
+          name: 'service',
+          value: [this.jurisdictions[0]]
```
udaydenduluri33, line 224: Why is it always the first jurisdiction?

udaydenduluri33 (approved):

Surekha-HMCTS (approved):

PR #1469, opened 2021-11-15
Feature Global Search

johnbenjamin-hmcts:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/EUI-4626
https://tools.hmcts.net/jira/browse/EUI-4396
https://tools.hmcts.net/jira/browse/EUI-4371
https://tools.hmcts.net/jira/browse/EUI-4595
https://tools.hmcts.net/jira/browse/EUI-4733

### Change description ###
All Global Search related changes

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #2692, opened 2023-01-20
Resolved issue with multiple subscriptions

phillip-whitaker-hmcts:

https://tools.hmcts.net/jira/browse/EUI-7539

### Change description ###

This bug was introduced by the following change in Global Search, which went into WA-R3 [here](https://github.com/hmcts/rpx-xui-webapp/pull/1469/commits/f1ab4abd60293b5f3e79e41d98f74692a278fe8f#diff-e73a2f86e3f032419620978f9907d8fd4bd6f59c00c401688cd2b1db7c062451L53). 

This simple addition of loading the user details inside of the header component had a knock-on effect on the app.component which also dispatches and subscribes to new changes of the UserDetails. 

The app component subscribes to the user details and collects information about the user timeout and then setups the TimeoutNotificationService, but also calls [this](https://github.com/hmcts/rpx-xui-webapp/blob/master/src/app/containers/app/app.component.ts#L122) which subscribes to the notificationOnChange [here](https://github.com/hmcts/rpx-xui-webapp/blob/master/src/app/containers/app/app.component.ts#L196). This creates a NEW observable in the common-lib and returns it. 

What appears to be happening is each time the notification service onNotificationChange is called, it creates and returns a NEW observable from the event emitter. This then causes, in some cases for the events to be missed. So when the change was made to add a new dispatch of the user details in the header, it caused this code and in app.component to be called again to setup the timeout service. And also creates a new subscription to a newly created observable. I guess because the common-lib publishes to the old observable and not the newly created one, or visa-versa. Which explains why it was so intermittent, because in part there was a race-condition to this. 

The change I have made is to subscribe to notificationOnChange ONCE and then subscribe to that observable, rather than asking common-lib to generate a new one each time. 

```
[ ] Yes
[x] No
```


RiteshHMCTS (dismissed):

RiteshHMCTS (approved):

src/app/containers/app/app.component.ts
```diff
@@ -3,7 +3,7 @@ import { Title } from '@angular/platform-browser';
 import { Router, RoutesRecognized } from '@angular/router';
 import { CookieService, FeatureToggleService, FeatureUser, GoogleTagManagerService, TimeoutNotificationsService } from '@hmcts/rpx-xui-common-lib';
 import { select, Store } from '@ngrx/store';
-import { combineLatest, Subscription } from 'rxjs';
+import { combineLatest, Observable, Subscription } from 'rxjs';
```
GaryScottHMCTS, line 6: Observable not being used

Mo-Lala84 (approved):
approved

src/app/containers/app/app.component.ts
```diff
@@ -315,7 +304,11 @@ export class AppComponent implements OnInit, OnDestroy {
       idleServiceName: 'idleSession'
     };
 
+    this.timeoutNotificationsService.notificationOnChange().subscribe(event => {
```
GaryScottHMCTS, line 307: would a take(1) work here so we can unsubscribe as we are finished with the event/subscription?

src/app/containers/app/app.component.ts
```diff
@@ -299,6 +287,7 @@ export class AppComponent implements OnInit, OnDestroy {
    * @param totalIdleTime - Should reach here in minutes
    */
   public initTimeoutNotificationService(idleModalDisplayTime, totalIdleTime) {
+    if (this.timeoutNotificationServiceInitialised) return;
```
GaryScottHMCTS, line 290: is this common to do shorthand if statements? Personally not a fan. But leave it if its common practice

src/app/containers/app/app.component.ts
phillip-whitaker-hmcts: No it wouldn't work in this case, I did try that

PR #2784, opened 2023-02-17
enable the new Task model refined to true

udaydenduluri33:

https://tools.hmcts.net/jira/browse/EUI-7837

### Change description ###

Enable new model for the feature toggle 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/work-allocation/services/work-allocation-case.service.ts
+++ b/src/work-allocation/services/work-allocation-case.service.ts
@@ -47,7 +47,7 @@ export class WorkAllocationCaseService {
   }
 
   public getTasksByCaseId(caseId: string): Observable<Task[]> {
-    return this.http.post<Task[]>(`${BASE_URL}/task/${caseId}`, {refined: false});
+    return this.http.post<Task[]>(`${BASE_URL}/task/${caseId}`, {refined: true});
   }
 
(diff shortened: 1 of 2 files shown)


connorpgpmcelroy (dismissed):

DavidJayakumar (approved):

Surekha-HMCTS (approved):

src/work-allocation/services/work-allocation-case.service.ts
```diff
@@ -47,7 +47,7 @@ export class WorkAllocationCaseService {
   }
 
   public getTasksByCaseId(caseId: string): Observable<Task[]> {
-    return this.http.post<Task[]>(`${BASE_URL}/task/${caseId}`, {refined: false});
+    return this.http.post<Task[]>(`${BASE_URL}/task/${caseId}`, {refined: true});
```
connorpgpmcelroy, line 50: This will defeat the purpose of the refined field, basically meaning new permission logic is used every time (see below for what we used to use for this). Is this what we want?

udaydenduluri33: Yes please, thats fine, If you see the discussion. From now on we will always enable it as true

connorpgpmcelroy (approved):
Approved after Uday's comment

PR #2843, opened 2023-03-08
Feature/e UI 6406 newer

MateiHMCTS:
https://tools.hmcts.net/jira/browse/EUI-6406

### Change description ###
Manage User - Update user details

```
[ ] Yes
[x] No
```


api/staff-ref-data/mock-staff-ref-data-api.ts
```diff
@@ -91,9 +91,8 @@ export class MockStaffRefDataAPI implements StaffRefDataAPI {
     }
   }
 
-  async getStaffRefUserDetails(req, res: Response, next: NextFunction) {
-    const id = req.params.id;
-    const apiPath: string = `/refdata/case-worker/user-details/${id}`;
+  async getStaffRefUsersById(req, res: Response, next: NextFunction) {
```
connorpgpmcelroy, line 94: Why is this getStaffRefUsersById if there isn't an ID?

api/staff-ref-data/real-staff-ref-data-api.ts
```diff
@@ -189,17 +189,28 @@ export class RealStaffRefDataAPI implements StaffRefDataAPI {
...
     workarea.forEach(workArea => {
       services.push(workArea.area_of_work);
-    })
+    });
     return services;
   }
 
-  async updateUserStatus(req, res, next: NextFunction) {
-    const id = req.params.id;
+  async getStaffRefUsersById(req, res, next: NextFunction) {
```
connorpgpmcelroy, line 196: Same question as above

connorpgpmcelroy (changes requested):
Query regarding fetchUsersById - Looks like it's not using ID in node layer?

api/staff-ref-data/mock-staff-ref-data-api.ts
MateiHMCTS: We need to pass in an array of email_ids. it's in req.body

connorpgpmcelroy (approved):
Would prefer we change the name so that it is not 'getStaffUsers...' because it is not a get request - find would be better wording

DavidJayakumar (approved):

PR #2860, opened 2023-03-14
WIP: EUI-4129: Activity Tracker

LucaDelBuonoHMCTS:
https://tools.hmcts.net/jira/browse/EUI-4129

### Change description ###
WebSocket Activity Tracker

```
[ ] Yes
[X] No
```


dean-bullock (dismissed):
Approved. Lots of formatting changes make the diff quite noisy. Something to look at getting everyone's code editor formatters in-line

Mo-Lala84 (dismissed):
approved

bradley-hmcts (dismissed):

api/application.ts
```diff
@@ -58,6 +58,76 @@ export async function createApp() {
     app.use(helmet.noSniff());
     app.use(helmet.frameguard({ action: 'deny' }));
     app.use(helmet.referrerPolicy({ policy: ['origin'] }));
+    app.use(helmet.hidePoweredBy());
+    app.use(helmet.hsts({ maxAge: 28800000 }));
+    app.use(helmet.xssFilter());
+    app.use(helmet.contentSecurityPolicy({
```
connorpgpmcelroy: This is no longer needed and should not be necessary - you should re-evaluate this file because I don't think any of the above is necessary unless it is relevant for activity tracker (which I don't think it is)

src/cases/resolvers/activity.resolver.ts
```diff
@@ -1,18 +1,28 @@
...
+  }
 
   public resolve(): Observable<boolean> {
-    try {
-      this.activityService.verifyUserIsAuthorized();
-    } catch (err) {
-      console.log(err);
-    }
-    return of(true);
+    return of(this.activityService.mode !== ActivityService.MODES.off);
```
connorpgpmcelroy: This boolean is referred to twice - should be a boolean variable

connorpgpmcelroy (changes requested):
Needs a few changes

src/work-allocation/components/case-name-field/case-name-field.component.ts
```diff
@@ -21,10 +22,10 @@ export class CaseNameFieldComponent implements OnChanges {
   public ngOnChanges(): void {
     let href: string;
     if (this.caseId) {
-      const jurisdictionId = this.jurisdiction;
-      const caseTypeId = this.caseType;
-      const caseId = this.caseId;
-      if (caseId && jurisdictionId && caseTypeId) {
+      if (this.caseId && this.jurisdiction && this.caseType) {
```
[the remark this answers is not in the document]
balajisridharanhmcts: Done

src/cases/resolvers/activity.resolver.ts
connorpgpmcelroy: Hasn't actually been replaced here though right? Should be return of(this.isActivityEnabled())?

src/work-allocation/components/case-reference-field/case-reference-field.component.ts
```diff
@@ -19,13 +19,16 @@ export class CaseReferenceFieldComponent implements OnChanges {
...
+    let caseType = this.caseType;
+    let caseReference = this.caseReference;
+    if (caseReference) {
       // Get rid of the spaces in the caseReference.
-      if (this.caseReference && this.jurisdiction && this.caseType) {
-        const caseId = encodeURIComponent(this.caseReference);
-        const jurisdictionId = encodeURIComponent(this.jurisdiction);
-        const caseTypeId = encodeURIComponent(this.caseType);
-        href = `${AppConstants.CASE_DETAILS_URL}${jurisdictionId}/${caseTypeId}/${caseId}`;
+      if (caseReference && jurisdiction && caseType) {
```
connorpgpmcelroy: Same as above, don't need two ifs here

connorpgpmcelroy (dismissed):
Approving because my concerns are now minor and I won't be here for the second half of the day to change my review, however would ideally like those comments addressed

PR #3059, opened 2023-06-06
Feature/e UI 8379

MateiHMCTS:
https://tools.hmcts.net/jira/browse/EUI-8379

### Change description ###
Document uploader in write query views
Refactor to avoid duplication

```
[ ] Yes
[X] No
```


DavidJayakumar (approved):
LGTM

src/cases/cases.module.ts
```diff
@@ -70,6 +69,7 @@ import { CreateCaseEventTriggerResolver } from './resolvers/create-case-event-tr
 // from services
 import * as fromServices from './services';
 import { effects, reducers } from './store';
+import { RpxTranslationModule } from 'rpx-xui-translation';
```
Daniel-Lam, line 72: @MateiHMCTS - please re-order the imports

Daniel-Lam (changes requested):
@MateiHMCTS - just one minor change needed; otherwise OK

src/cases/cases.module.ts
MateiHMCTS: I think it's already ordered as '.' is before 'r'

Daniel-Lam: Hmm... interesting. In VSCode, for example, `rxjs` comes before `./users.effects` - can leave for now although the pipeline linter may flag it

MateiHMCTS: Nothing from yarn lint

Daniel-Lam (approved):

PR #3120, opened 2023-06-28
Use supported version of yarn

timja:
Node 14 went out of support in October 2021

Going to run this through the pipeline and see if all ok


timja:
Angular needs updating:
https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3

PR #3186, opened 2023-07-27
EXUI-661 - Fix CSP config

danlysiak:
https://tools.hmcts.net/jira/browse/EXUI-661

### Change description ###
Fix CSP for GA

```
[ ] Yes
[x] No
```


config/default.json
```diff
@@ -197,16 +197,18 @@
           "'self'",
```
danlysiak, line 197: Suspect the changes in this file are not needed, as the config is via Helmet in the `application.ts`, but changing here anyway just in case and for consistency.

RiteshHMCTS (approved):

codaimaster (approved):
lgtm

mwaheedanjum (approved):

PR #3317, opened 2023-10-10
EUI-8460 QM Raise a query submission

johnbenjamin-hmcts:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/EUI-8460

### Change description ###
Query management - Raise a query submission

### Checklist


src/app/models/user-details.model.ts
```diff
@@ -2,6 +2,7 @@ export interface UserInfo {
   id: string;
   forename: string;
   surname: string;
+  name?: string;
```
andywilkinshmcts, line 5: We should have a chat about this. If we can avoid it, I'd really like to, because it's going to cause a lot of confusion in future I imagine this is being driven by the fact that "CaseMessage" has only a name field. How about if we adopt the convention of using the surname field for name?

andywilkinshmcts (changes requested):

src/app/models/user-details.model.ts
olusegz07: Property name is currently not required.  It has been removed

andywilkinshmcts (approved):

RiteshHMCTS (approved):

PR #3357, opened 2023-11-06
exui-969-hide-hearing-tab-for-solicitor

RiteshHMCTS:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/EXUI-969

### Change description ###
hide-hearing-tab-for-solicitor

### Checklist


olusegz07 (dismissed):

OgunyemiO (dismissed):

src/cases/utils/utils.ts
```diff
@@ -61,7 +62,7 @@ export class Utils {
     if (featureVariation.jurisdiction === jurisdictionId && featureVariation.caseType === caseType) {
       if (userRoles) {
         return featureVariation.roles ? userRoles.some((userRole) =>
-          featureVariation.roles.some((role) => role === userRole)) : false;
+          featureVariation.roles.some((role) => role === userRole)) && !userRoles.includes(PUI_CASE_MANAGER) : false;
```
andywilkinshmcts, line 65: I don't like this fix, because you've done the test inside the outer some function, which means it has to do the userRoles.Includes for each role listed in the variation

It doesn't make much practical difference, and I think what you've done will probably work, so this is probably nit picking,  but I think what you want is :
return !userRoles.includes(PUI_CASE_MANAGER) && featureVariation.roles ? userRoles.some((userRole) =>
          featureVariation.roles.some((role) => role === userRole)) : false;
          
That way it will fail at once if the user is a solicitor

andywilkinshmcts (changes requested):
Please re-do the check for the solicitor role as I've suggested in my comment. It doesn't make a huge amount of difference, so if what I've suggested doesn't immediately work it's ok to go with the current fix

src/cases/utils/utils.ts
RiteshHMCTS: done the changes as mentioned

PR #3379, opened 2023-11-17
cot-599: Reasonable Adjustments incorrectly showing as amended

richardhenryash:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/COT-599

### Change description ###
Fix for bug COT-599. Added and implemented reasonable adjustments amended converter and unit tests.

### Checklist


src/hearings/converters/reasonable-adjustments.amended.converter.ts
```diff
@@ -0,0 +1,28 @@
...
+export class ReasonableAdjustmentsAmendedConverter implements IsAmendedConverter {
+  public transformIsAmended(hearingState$?: Observable<State>): Observable<boolean> {
+    return hearingState$.pipe(map((state) => {
+      return this.checkReasonableAdjustments(state.hearingRequest.hearingRequestMainModel.partyDetails,
+        state.hearingRequestToCompare.hearingRequestMainModel.partyDetails);
+    }));
+  }
+
+  public checkReasonableAdjustments(originalPartyDetails: PartyDetailsModel[], partyDetailsToCompare: PartyDetailsModel[]) {
+    return partyDetailsToCompare.some((party) => {
```
andywilkinshmcts, line 18: I'm not sure that some() is appropriate here. some will look at the parties until it finds one that matches, and then it will stop. I think you probably want every() rather than some, and you want a test case with say 3 parties, where the first 2 match and the 3rd doesn't

andywilkinshmcts (changes requested):
every, not some?

andywilkinshmcts (approved):

src/hearings/converters/reasonable-adjustments.amended.converter.ts
richardhenryash: Hi Andy, the function checkReasonableAdjustments checks if the reasonable adjustments are different between the parties (it will return true for the first party it finds with different reasonable adjustments and stop processing at that point).

PR #3393, opened 2023-11-30
GA-127 Add script for local config setup

shaed-parkar:
### Jira link (if applicable)

[GA-127](https://tools.hmcts.net/jira/browse/GA-127)

### Change description ###

Add a setup folder to get the configuration needed to run the web app locally


ben-james-hmcts (approved):

RiteshHMCTS (dismissed):

andywilkinshmcts (changes requested):
I'm OK with you doing this, though do you really need the script that dumps the secrets? They're called secrets for a reason, and I'm not so comfortable with a tool that lets them be dumped, especially if it can be pointed at prod. If you only need it work on AAT, I'd prefer it it was hard coded in the script to only work on AAT - basically move the REMOTE_ENV variable into the script. I don't want the secrets ending up in a log file anywhere

danlysiak:
> > I'm OK with you doing this, though do you really need the script that dumps the secrets? They're called secrets for a reason, and I'm not so comfortable with a tool that lets them be dumped, especially if it can be pointed at prod. If you only need it work on AAT, I'd prefer it it was hard coded in the script to only work on AAT - basically move the REMOTE_ENV variable into the script. I don't want the secrets ending up in a log file anywhere
> 
> I appreciate the concern however, the script to retrieve the current secrets is a much safer approach than passing config files over IM/Email. A semi decent access policy should limit access to the prod key vault. This script is also no different than manually going into a key vault and copying the values locally, except how much time is spent doing it.

Developers don't have read access to secrets in Prod anyway, only non-Prod 👍

RiteshHMCTS (approved):

Josh-HMCTS (approved):

andywilkinshmcts (changes requested):
I don't want this code in the repo. I don't see anything that prevents it from being deployed to production, which might result in a file with the secrets being written to the production server, which might theoretically be retrieveable via a browser

sonny-hmcts:
I am not sure if this helps but I created a small cli application for downloading and mounting the secrets to the /mnt/secrets directory on the local development machine (mac only at this time). 

the property-volumes library used by this project will by default attempt to load the secrets from the /mnt/secrets folder. 

the cli tool works with both pre and post Catalina mac os and can be found here [pvmount-tool](https://github.com/sonny-hmcts/pvmount-tool ). it can be built into an executable and installed so the command is always available in the terminal. 

in addition to this if you are working with a docker container then simply mount the /mnt/secret folder on the local machine to the /mnt/secret folder inside the container.

having a tool outside of the project bypasses the need for a script to be contained in the project satisfying Andy's concern of the script being run while on a production server.

PR #3398, opened 2023-12-05
Change default environment to point to production

richardhenryash:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/COT-639

### Change description ###
Change default environment to point to production

### Checklist


andywilkinshmcts (dismissed):
we need this change to be made, but we should be careful about merging it, because it might break AAT. We need to get to the bottom of the issue with the LOV api config that you have been looking at. I'm going to mark this as do not merge for now, but we can remove that label when we understand the problem

PR #3442, opened 2024-01-12
hearing button loading  fix

OgunyemiO:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/EXUI-1192

### Change description ###
Fixes "Request Hearing" button early loading error

### Checklist


richardhenryash (dismissed):

src/cases/containers/case-hearings/case-hearings.component.ts
```diff
@@ -141,6 +142,7 @@ export class CaseHearingsComponent implements OnInit, OnDestroy {
             );
           }
           const caseHearingViewModels: HearingListViewModel[] = this.calculateEarliestHearingDate(caseHearingModels);
+          this.showLoadSpinner = false;
```
andywilkinshmcts, line 145: This feels like quite a random place to do this. We talked about doing it at the end of the ngOnInit - was there a reason for it being here?

src/cases/containers/case-hearings/case-hearings.component.html
```diff
@@ -1,4 +1,4 @@
-<xuilib-loading-spinner *ngIf="showSpinner$ | async"></xuilib-loading-spinner>
```
andywilkinshmcts, line 1: I don't think this is  a good idea. The  loading service is tracking the loading of the service hearing values call, and this is asynchronous. You are now depending on it to complete in a timely way before the spinner is pulled down by the LSH call completing. Probably it will, but we don't have control over how long it takes. If the service is calling CCD to fetch the data it might take a long time sometimes.

It might be better to have 2 observables - the existing showSpinner$ and a new one that emits at the end of ngOnInit, which are combined using combineLatest

andywilkinshmcts (changes requested):
I think the spinner code needs to remain asynchronous

src/cases/containers/case-hearings/case-hearings.component.html
OgunyemiO: Hi @andywilkinshmcts , I tried what we proposed on Friday but the problem persisted.  The solution above worked as it was put at the end of the function call which itself was called in the ngOnInit.

I will revisit the ticket to address your concerns though.

OgunyemiO: @andywilkinshmcts , please check my latest fixes. Thanks.

richardhenryash (dismissed):

richardhenryash (approved):

olusegz07 (dismissed):

Josh-HMCTS (approved):

connorpgpmcelroy (approved):
LGTM

RiteshHMCTS (dismissed):

andywilkinshmcts (changes requested):
Seems pretty hacky. It probably works, but I don't like the fact that you've introduced an extra observable - surely it should be possible to fix this without this complication. The displaySpinner property shouldn't be needed either. You should be able to 
<xuilib-loading-spinner *ngIf="showSpinner$ | async" > </xuilib-loading-spinner>

src/cases/containers/case-hearings/case-hearings.component.html
```diff
@@ -50,6 +53,6 @@
         [actions]="hearingsActions">
       </exui-case-hearings-list>
     </ng-container>
-
+  </ng-container>
   </ng-container>
```
connorpgpmcelroy, line 57: Minor comment - Indentation is off here

connorpgpmcelroy (approved):
Approved with minor comment

RiteshHMCTS (approved):

anthonydummer (dismissed):
LGTM

RiteshHMCTS (approved):

kasi-subramaniam (approved):

PR #3446, opened 2024-01-15
rate limit for document uploads added

olusegz07:
### Checklist


api/documents/index.ts
```diff
@@ -0,0 +1,47 @@
+// Time configuration
+const INITIAL_TIMEOUT_PERIOD = 5000; // 5 seconds
+const MAX_TIMEOUT_PERIOD = 180000; // 180 seconds
+
+let isRequestRateLimited = false;
```
andywilkinshmcts, line 5: You can't do this - it's a global variable. Needs to be a session variable

api/documents/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  return true;
+}
+
+// Handle responses being sent back to the client
+export function handleResponse(req, res, proxyRes, json) {
+  // Current time stored as the last time a document upload was completed
+  res.session.lastUploadTime = Date.now();
+
+  // Double the timeout period up to the maximum, if not rate-limited
+  if (!isRequestRateLimited) {
```
andywilkinshmcts, line 37: If we get another request while we are within the rate limit timeout, then double the timeout, up to some max value. The isRequestRateLimited value (which needs to be session specific, not global) is set to true if we get a request within the timeout period. Therefore I think this needs to be inverted - it should say:
 if (req.session.isRequestRateLimited)

api/documents/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    const nextTimeout = (res.session.nextTimeout || INITIAL_TIMEOUT_PERIOD) * 2;
+    res.session.nextTimeout = Math.min(nextTimeout, MAX_TIMEOUT_PERIOD);
+  } else {
+    delete res.session.nextTimeout;
```
andywilkinshmcts, line 41: Not sure that you want to do this - should be set to the default timeout

andywilkinshmcts (changes requested):

api/documents/index.ts
olusegz07: As discussed, this is valid, as code in line 15 addresses the concern raised.

  // If there is no timeout period in the session, set it to the default
  const timeoutPeriod = nextTimeout || defaultTimeoutPeriod;

OgunyemiO (approved):

andywilkinshmcts (dismissed):

andywilkinshmcts (dismissed):

OgunyemiO (dismissed):

PR #3474, opened 2024-01-19
Added unique tab text to document viewer

OgunyemiO:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/EXUI-190

### Change description ###
On the current system(not CCD or JCM) the judges  tend to have multiple tabs open at the same time for all the different cases they are working on during the day, They want the name of each tab to be unique so that the user has an idea of what case is on what tab to ease navigation.

### Checklist

### DIFF
--- a/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
+++ b/src/app/components/media-viewer-wrapper/media-viewer-wrapper.component.ts
@@ -3,6 +3,7 @@ import { WindowService } from '@hmcts/ccd-case-ui-toolkit';
 import { FeatureToggleService } from '@hmcts/rpx-xui-common-lib';
 import { Observable } from 'rxjs';
 import { SessionStorageService } from '../../services/session-storage/session-storage.service';
+import { Title } from '@angular/platform-browser';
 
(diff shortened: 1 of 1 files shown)


RiteshHMCTS (changes requested):
@OgunyemiO can you add unit test to cover this change

andywilkinshmcts (changes requested):
I am skeptical that the tests for this component will still work, because you've injected another service in the constructor. I think you need to create a mock object and inject that as is done for other services in the spec file. Ideally there would be a test that the when you load a document the window title gets set too.

richardhenryash (approved):

richardhenryash (approved):

RiteshHMCTS (approved):

olusegz07 (approved):

RiteshHMCTS (dismissed):

olusegz07 (dismissed):

andywilkinshmcts (approved):
You'll get a conflict if you try and merge on the toolkit version, but the actual change is fine

RiteshHMCTS (approved):

kasi-subramaniam (approved):

PR #3552, opened 2024-02-26
COT-777: When hearing day is not required than hearing timing and participants shouldn't be displayed

SomeshRA:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/COT-777

### Change description ###
When hearing day is not required than hearing timing and participants shouldn't be displayed


andywilkinshmcts (dismissed):
Try and avoid making extensive whitespace only changes - you may need to configure your IDE to stop it automatically reformatting files. Is this a complete fix for all the items in the ticket?

RiteshHMCTS (dismissed):

andywilkinshmcts (approved):

olusegz07 (approved):

PR #3569, opened 2024-03-04
Task/1194 security issue

olusegz07:
### Checklist


RiteshHMCTS (approved):

OgunyemiO (dismissed):

OgunyemiO (dismissed):

RiteshHMCTS (approved):

andywilkinshmcts (approved):

api/documents/index.ts
```diff
@@ -0,0 +1,42 @@
+// Time configuration
+const INITIAL_TIMEOUT_PERIOD = 5000; // 5 seconds
```
GauravChawlaMOJ, line 2: In the ticket says, it should be configurable but it is hardcoded here. Imo this is not a user friendly feature more so when users are not aware of this 5sec timeout. There's no message shown which informs user about this if the doc upload fails for them. 
 From the ticket - _The initial timeout parameter (5 seconds) should be configurable_

PR #3664, opened 2024-05-02
Make caseworker session storage change

connorpgpmcelroy:
Use node caching while still waiting for API to cache

### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/EXUI-1772

### Change description ###
Replace caseworker session storage on angular layer with caching on node layer (to be replaced by API caching eventually). Not that we also cache users after role category has been populated by role assignment which will not be cached by API so some caching will be kept


api/workAllocation/caseWorkerService.ts
```diff
@@ -6,7 +6,14 @@ import { setHeaders } from '../lib/proxy';
 import { CaseworkerPayload, ServiceCaseworkerData } from './interfaces/caseworkerPayload';
 
 const logger: JUILogger = log4jui.getLogger('caseworker-service');
-const MAX_RECORDS: number = 10000;
+const MAX_RECORDS: number = 100000;
```
andywilkinshmcts, line 9: Is there evidence that you need to go to 100K? Can we go to 20K or 30K initially. Another option is to make multiple calls for 10K - API supports pagination.

api/workAllocation/caseWorkerUserDataCacheService.ts
```diff
@@ -0,0 +1,94 @@
...
+import { ServiceUserDetailList, StaffUserDetails } from './interfaces/staffUserDetails';
+import { EnhancedRequest } from '../lib/models';
+import { getWASupportedJurisdictionsList } from '../waSupportedJurisdictions';
+import { mapUsersToCaseworkers, prepareGetUsersUrl, prepareRoleApiRequest, prepareRoleApiUrl } from './util';
+import { handlePostRoleAssignments, handleUsersGet } from './caseWorkerService';
+import { NextFunction } from 'express';
+import { Caseworker } from './interfaces/common';
+import { FullUserDetailCache } from './fullUserDetailCache';
+
+const TTL = 60;
```
andywilkinshmcts, line 11: 60 sec TTL seems quite aggressive - maybe 5 mins?

api/workAllocation/fullUserDetailCache.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class FullUserDetailCache {
+  // singleton class to keep single instance of role assignment query
+  private static instance: FullUserDetailCache;
+  private userDetails: Caseworker[];
+
+  static getInstance(): FullUserDetailCache {
+    if (this.instance) {
+      return this.instance;
+    }
```
andywilkinshmcts, line 11: Hooray for single threaded languages

andywilkinshmcts (approved):
Very comprehensive job. I'm a bit worried about a couple of parameters - TTL and number of records fetched - let's discuss

api/workAllocation/caseWorkerService.ts
connorpgpmcelroy: As discussed on standup, the current number is suitable for the moment but may be looked at in future

Josh-HMCTS (dismissed):

RiteshHMCTS (approved):

PR #3668, opened 2024-05-07
Ex UI 1856 present button for document in case file view on aat

OgunyemiO:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/EXUI-1856

### Checklist


RiteshHMCTS (approved):

andywilkinshmcts (commented):
Doesn't need to be reviewed because it has a pre-release version of the toolkit and so can't be merged. What needs to happen is that we need to merge the toolkit change, which can be done without SRT and then update this PR to point to that version.

OgunyemiO:
> Doesn't need to be reviewed because it has a pre-release version of the toolkit and so can't be merged. What needs to happen is that we need to merge the toolkit change, which can be done without SRT and then update this PR to point to that version.

@andywilkinshmcts ,  I think we need to merge, in this particular instance. There are few configuration file changes in MC that compliment the changes in Toolkit. Otherwise, the changes in toolkit won't reflect/work.

PR #3692, opened 2024-05-23
DTSSE-5011 - update playwrights version to 1.42

david-gimelle:
https://tools.hmcts.net/jira/browse/DTSSE-5011

add playwright/test  lib and update playwright lib to version 1.42


david-gimelle:
no needed

PR #3703, opened 2024-05-29
DTSSE 5018 - login and log out from EUI with valid user

david-gimelle:
https://tools.hmcts.net/jira/browse/DTSSE-5018)

Rewriting of login and log out from EUI with valid user as a pure playwright test


README.md
```diff
@@ -180,7 +180,7 @@ Run `ng e2e` to execute the end-to-end tests via [Protractor](http://www.protrac
 
 ## Running pure playwright end-to-end tests
 
-Run `HEAD=true TEST_URL=<testurl> yarn test:fullfunctional2` to execute the pure playwright end-to-end tests via [Playwright](https://playwright.dev/).
+Run `HEAD=true CASEAPIBASEURL=https://manage-case.aat.platform.hmcts.net/cases yarn test:fullfunctional2` to execute the pure playwright end-to-end tests on aat via [Playwright](https://playwright.dev/).
```
andywilkinshmcts, line 183: It's very minor, but it annoys me slightly that the env variable is called CASEAPIBASEURL, since there isn't really an API on that URL. Could we call it APP_URL or something? Doesn't really matter if it's a nuisance

playwright_tests/E2E/login-and-log-out-with-valid user.test.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/test';
+import config from "../config"
+const testConfig = require('../../test_codecept/e2e/config/appTestConfig');
+
+test('login and log out from EUI with valid user', async ({ page }) => {
+  const matchingUsers = testConfig.users[testConfig.testEnv].filter((user) => user.userIdentifier === 'PROD_LIKE');
+  await page.goto(config.CaseAPIBaseURL);
```
andywilkinshmcts, line 7: Actually I really want you to change it now that it's become an attribute. Let's call it AppBaseURL

andywilkinshmcts (changes requested):

README.md
david-gimelle: Done

playwright_tests/E2E/login-and-log-out-with-valid user.test.ts
david-gimelle: Done

PR #3731, opened 2024-06-17
Add Global Search Tests

david-gimelle:
https://tools.hmcts.net/jira/browse/DTSSE-5120

Mirror Global Search Test Scenario as pure playwright scenario


david-gimelle:
not needed

PR #3738, opened 2024-06-19
DTSSE 5120 - Case flags tab

david-gimelle:
https://tools.hmcts.net/jira/browse/DTSSE-5120

Mirror Case flags tab Scenario as pure playwright test


playwright_tests/E2E/case-flags.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  const tableName = "Applicant";
+  await checkNumberOfRow(page, tableClass, tableName, 5);//One row for the header
+  await expect(page.getByRole('table', { name: 'Applicant' }).getByRole('caption')).toBeVisible();
+  
+  const rowTextRef = "Support filling in forms";
+  const textsToCheck = [/Test/, /24 Oct 2023/, /Active/];
+  await checkTableRowContent(page, tableClass, tableName, rowTextRef, textsToCheck);
+
+  console.log("Check line 1");
+  await checkTableCellContent(page, tableName, 1, 0, "Documents in a specified colour");
```
andywilkinshmcts, line 43: I think it would be worth writing a simple function to handle tables of content like:
interface TestDataRow{
  column: number,
  row: number,
  testData: string 
};
function checkTable(page: number, tableName: string, data: TestDataRow[]) {
  data.forEach (d => await checkTableCellContent(page, tableName, d.column, d.row, d.testData))
}

then you could replace n calls to checkTableCellContent with 

checkTable (page, tableName, [ 
{ 1, 1, "Test" },
{ 1, 2, "24 Oct 2023 }
 ...
 {2, 0, "Documents in a specified colour"}
etc
 }
It's not much effort, makes things a little more compact, and potentially allows for parallelisation of the checks in future. There's also potential to locate the tablename once and pass it in rather than fetching it every time, which should speed things up

andywilkinshmcts (approved):
looks good, though there is potential for some refactoring to make it more compact and efficient

PR #3791, opened 2024-07-09
DTSSE-5202 - Faster playwright config

david-gimelle:
https://tools.hmcts.net/jira/browse/DTSSE-5202
Update Playwright config for a faster end 2 end tests. Notice Playwright E2E tests don't run in any pipeline for now.


david-gimelle:
not needed anymore

PR #3833, opened 2024-08-09
EXUI_2193: Local mc-menu-item config

Josh-HMCTS:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/EXUI-2193

### Change description ###

Move mc-menu-items out of LD


andywilkinshmcts (changes requested):
this looks like it will work, but you might have missed an opportunity with the various environment configs
I think they should all be defined in terms of the prod config, something like

aatConfig = {
  ...prodConfig,
  // differences only go here
 }
  
This should be easier to maintain, though in some cases it's a bit fiddly
Can you see if it's practical to capture the differences between AAT and Prod in this way?

api/configuration/menuConfigs/configs.ts
```diff
@@ -0,0 +1,57 @@
...
+import { aatDifferences } from './aat-diffs';
+import { previewDifferences } from './preview-diffs';
+
+export function setupMenuConfig(environment) {
+  let config;
+  if (environment === 'prod') {
+    config = baseConfig;
+  } else if (environment === 'aat') {
+    let aatConfig = { ...baseConfig };
+    const rolesRegexChanges = {
```
andywilkinshmcts, line 11: This bit is just a screw up really. It should be (judge)|(judiciary)|(panelmember) in all configs
    }

andywilkinshmcts (commented):
I like it. Quite clear what is what

api/configuration/menuConfigs/configs.ts
andywilkinshmcts: You can leave it for now if you like, or if you want to you can just eliminate that difference

andywilkinshmcts (dismissed):

RiteshHMCTS (dismissed):

Josh-HMCTS:
This has been merged in to the SRT branch

PR #3834, opened 2024-08-12
Add cherrypicked axe-core accessibility tests

hemantt:
### Checklist


david-gimelle (dismissed):
All good for me.

andywilkinshmcts (changes requested):
I think you should do something a little smarter about enabling the axeTests. In the short term I would uncomment all the calls to "axeTest" and within that function have some code like this 
if (!process.env?.ENABLE_AXE_TESTS)
  return

PR #3850, opened 2024-08-22
EXUI-2277 - Move mc-menu-theme out of LD 

Josh-HMCTS:
### Jira link

See [EXUI-2277](https://tools.hmcts.net/jira/browse/EXUI-2277)

### Change description

Move the mc-menu-theme out of launch darkly and in to angular config


src/app/containers/app-header/app-header.component.ts
```diff
@@ -128,17 +121,22 @@ export class AppHeaderComponent implements OnInit, OnDestroy {
...
-  }
-
-  public logErrorAndReturnDefaultTheme(error): Observable<ApplicationTheme> {
-    this.loggerService.error(error);
-    return of(this.defaultTheme);
+  public setApplicationThemeForUser(): void {
+    const availableThemes = environment.themes;
+    const userRolesSet = new Set(this.userRoles);
+    const userTheme = Object.keys(availableThemes).find((theme) =>
+      theme !== 'default' && theme.split('|').some((role) => userRolesSet.has(role))
```
andywilkinshmcts, line 128: Can you make the role matching for themes a regex please. What you've done works, but isn't functionally equivalent to what we had. You should also change default to '.+'
Then you could do something like 
const userTheme = Object.keys(availableThemes.some((themeRegex) => userRolesSet.some((role) => role.matches(themeRegex)))

andywilkinshmcts (changes requested):

StoneBenHMCTS (dismissed):

andywilkinshmcts (dismissed):
Thanks - looks good

RiteshHMCTS (approved):

andywilkinshmcts (approved):

PR #3853, opened 2024-08-26
EXUI-2245 initial checkin

anthonydummer:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2245

### Change description

An error meant that the service hearing values were always being used.  These should only be used when in edit view mode and a change had been identified and not confirmed.  Code has been simplified to separate the identification of which data source to use and the population of that data. 

### Testing done

Testing the scenario on the ticket and other tickets showed that changes made and confirmed on the hearing request page were being reflected on the summary page, but on re-entry to the hearing timing page, these were being lost. 
Testing each of the values on the page, I have verified that any saved values are subsequently present when returning to the page. The service hearing values can also be seen on first entry to this page. 

### Checklist


StoneBenHMCTS (dismissed):

andywilkinshmcts (approved):
This approach seems sensible, though I now wonder if it shouldn't be used a lot more widely than in just this one component

RiteshHMCTS (dismissed):

andywilkinshmcts (approved):

RiteshHMCTS (dismissed):

andywilkinshmcts (approved):

RiteshHMCTS (approved):

Josh-HMCTS (approved):

PR #3877, opened 2024-09-06
Ex UI 2284

sonny-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2284
https://tools.hmcts.net/jira/browse/EXUI-2285

### Change description

added a LAU service with the aim to create an audit trail on requested access.
this includes new configurations for the API endpoint and a feature config to turn the feature on or off

### Testing done

Manual testing & automated

### Checklist


MartinYSpasov (approved):

api/services/lau.ts
```diff
@@ -0,0 +1,231 @@
...
+ * }
+ * @param req
+ */
+async function createAccessLogFromRequest(req: EnhancedRequest) {
+  const process = req.body.roleRequest.process;
+  try {
+    const detail: detailObject = JSON.parse(req.body.requestedRoles[0].notes[0].comment);
+    const accessLog: AccessLog = {
+      requestType: (process === 'specific-access') ? REQUEST_TYPE.SPECIFIC : REQUEST_TYPE.CHALLENGED,
+      caseRef: req.body.requestedRoles[0].attributes.caseId,
```
andywilkinshmcts, line 119: For resilience, I think it would be good to use optional chaining in all these attribute accesses. As it stands you'll get no log if any of the data is malformed - a partial log is better than none I think. So:
reg?.body?.requestedRoles[0]?.attributes?.caseId 
etc

andywilkinshmcts (changes requested):
Minor change to improve robustness, otherwise very neat and tidy

api/services/lau.ts
rapolaskaseliscgi: Addressed, though I didn't do for `req.body`, if we didn't have those, there wouldn't be anything to pass.

andywilkinshmcts (approved):

api/services/lau.ts
```diff
@@ -126,11 +126,10 @@ async function createAccessLogFromRequest(req: EnhancedRequest) {
       requestType: (process === 'specific-access') ? REQUEST_TYPE.SPECIFIC : REQUEST_TYPE.CHALLENGED,
       caseRef: requestedRole?.attributes?.caseId,
       userId: requestedRole?.actorId,
-      action: process === 'specific-access' ? ACTION_TYPE.CREATED : ACTION_TYPE.APPROVED,
+      action: process === 'specific-access' ? ACTION_TYPE.CREATED : ACTION_TYPE.AUTO_APPROVED,
       timestamp: requestedRole?.notes[0]?.time,
       reason: reason,
-      requestStartTimestamp: requestedRole?.notes[0]?.time,
-      requestEndTimestamp: requestedRole?.endTime
+      requestEndTimestamp: null
```
andywilkinshmcts, line 132: Why is the end timestamp null? It should be the end of the granted access

api/services/lau.spec.ts
```diff
@@ -176,7 +175,6 @@ describe('createAccessLog', () => {
       action: 'CREATED',
       timestamp: '2023-09-06T12:34:56.789Z',
       reason: 'mock-reason',
-      requestStartTimestamp: '2023-09-06T10:00:00.000Z',
```
andywilkinshmcts: I don't think you should remove this field. How will you represent when specific access is granted for a future period - e.g. for a week, in a month's time

andywilkinshmcts (changes requested):
I don't think you should be removing requestStartTime. You seem to be setting requestEndTime to null which I think is rarely, if ever, going to be correct

api/services/lau.ts
rapolaskaseliscgi: There is no way user can specify for how long he is requesting access. By default it is set to 1 month from now, which is misleading on LAU, because that is not what user requested

api/services/lau.spec.ts
rapolaskaseliscgi: Unless it is some future feature coming up, there isn't a way to ask or give permissions for specific timeframe in the future. The only option is to approve from now until the chosen time. And on LAU side it is always going to be "now" or even more confusingly "midnight of today (in the past)", even if permission was granted only during the day

MartinYSpasov:
Closing this PR because Jenkins piplines are not running the PR. Will open a new as temp solution to unblock e2e testing.

PR #3878, opened 2024-09-07
Danny - Webpack optimisations

furnivall:
This ticket brings in a few changes to the webpack config on the node api. Should in theory be able to speed things up a bit on the pipeline (or at least reduce the runtime of the underlying `yarn build:node` script). 

There may also be scope for smaller bundle sizes / runtime perf improvements too.


api/webpack-node.config.js
```diff
@@ -1,52 +1,65 @@
...
       new TerserPlugin({
+        parallel: true,
         terserOptions: {
-          sourceMap: true
-        }
-      })
-    ]
+          sourceMap: true,
+          compress: {
+            drop_console: true,
```
furnivall, line 18: I'm not 100% sure about whether this is a good idea but it does provide the biggest speedup of all these optimisations - @andywilkinshmcts what do you reckon?

PR #3902, opened 2024-09-18
EXUI-2179 - add call to get judicial user on restricted access case

Josh-HMCTS:
### Jira link

See [EXUI-2179](https://tools.hmcts.net/jira/browse/EXUI-2179)

### Change description

Added additional API call for restricted cases to load the judge users

### Testing done

Tested on Demo case found in ticket, now able to see 2 users in list


RiteshHMCTS (dismissed):

connorpgpmcelroy (approved):
LGTM

src/hearings/services/judicial-ref-data.service.ts
```diff
@@ -11,4 +11,9 @@ export class JudicialRefDataService {
     return this.http.post<JudicialUserModel[]>('api/prd/judicial/searchJudicialUserByPersonalCodes',
       { personal_code: personalCodes });
   }
+
+  public searchJudicialUserByIdamID(personalCodes: string[]): Observable<JudicialUserModel[]> {
```
andywilkinshmcts, line 15: this function is confused. personal codes are not idam ids - they are a completely different identifier. Check what you are passing to make sure that it actually is an idam ID, and rename the parameter

api/prd/judicial/index.ts
```diff
@@ -36,6 +36,23 @@ export async function searchJudicialUserByPersonalCodes(req: EnhancedRequest, re
   }
 }
 
+export async function searchJudicialUserByIdamId(req: EnhancedRequest, res: Response, next: NextFunction) {
+  const reqBody = req.body;
+  const markupPath: string = `${prdUrl}/refdata/judicial/users`;
+  try {
+    // Judicial User search API version to be used depends upon the config entry FEATURE_JRD_E_LINKS_V2_ENABLED's value
+    req.headers.accept = showFeature(FEATURE_JRD_E_LINKS_V2_ENABLED)
```
andywilkinshmcts, line 44: Good that you are accounting for this, but e links should be on everywhere now

src/cases/containers/restricted-case-access-container/restricted-case-access-container.component.ts
```diff
@@ -66,9 +68,30 @@ export class RestrictedCaseAccessContainerComponent implements OnInit, OnDestroy
 
   private getRestrictedCases(caseworkers: Caseworker[]): RestrictedCase[] {
     const restrictedCases: RestrictedCase[] = [];
-    this.idamIds.forEach((id) => {
+    this.idamIds.forEach(async (id) => {
       const user = caseworkers.find((caseworker) => caseworker.idamId === id);
       const caseRole = this.caseRoles.find((role) => role.actorId === id);
```
andywilkinshmcts, line 73: This code worries me a bit. It finds one role, and if that has the role category judicial then it searches. I think it should be checking if any role has the role category judicial, not just the first one with the right actor Id

andywilkinshmcts (changes requested):
A couple of areas need to be tightened up

src/hearings/services/judicial-ref-data.service.ts
Josh-HMCTS: Yeah this should have been renamed to idamIds, have updated

api/prd/judicial/index.ts
Josh-HMCTS: removed the feature check and just set to use v2 accept header

src/cases/containers/restricted-case-access-container/restricted-case-access-container.component.ts
Josh-HMCTS: I think this is a very poorly named variable as it really isn't caseRoles its an array of all the users that have access to the restricted case.
I have updated the variable name throughout this file to be usersWithAccess instead of caseRoles as i think that makes it clearer as to what the variable actually is.

andywilkinshmcts (approved):

PR #3909, opened 2024-09-24
Add regional-centre-admin

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2350

### Change description

Added regional-centre-admin to list of roles we look for when gathering role assignments

### Testing done

Verified relevant role was being returned from setting of role category to ADMIN


RiteshHMCTS (dismissed):

Josh-HMCTS (approved):

api/workAllocation/util.ts
```diff
@@ -366,7 +366,8 @@ export function prepareRoleApiRequest(jurisdictions: string[], locationId?: numb
     // TODO: This should not be hard-coded list
```
andywilkinshmcts, line 366: Can we move it to default.json and custom-environment-variables.json. You'll have to do a split on a comma separated value to get the array required here. I'm OK if you raise a separate tech debt ticket for this

andywilkinshmcts (approved):

PR #3910, opened 2024-09-25
Fix potential race condition in AppConfig init

andywilkinshmcts:
### Jira link
See [EXUI-2353](https://tools.hmcts.net/jira/browse/EXUI-2353)

### Change description
Fix race condition caused by observable not resolving in a timely fashion

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)
- [x] Does this PR introduce a breaking change

PR #3981, opened 2024-10-28
Initialise config with correct keys

danlysiak:
### Jira link

### Checklist

### DIFF
--- a/src/app/services/ccd-config/ccd-case.config.spec.ts
+++ b/src/app/services/ccd-config/ccd-case.config.spec.ts
@@ -160,7 +160,7 @@ describe('AppConfiguration', () => {
   }));
 
   it('should have getDocumentSecureMode return value', inject([AppConfig], (service: AppConfig) => {
-    expect(service.getDocumentSecureMode()).toBe(true);
+    expect(service.getDocumentSecureMode()).toBe(false);
   }));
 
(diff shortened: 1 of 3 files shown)


src/app/services/ccd-config/ccd-case.config.spec.ts
```diff
@@ -160,7 +160,7 @@ describe('AppConfiguration', () => {
   }));
 
   it('should have getDocumentSecureMode return value', inject([AppConfig], (service: AppConfig) => {
-    expect(service.getDocumentSecureMode()).toBe(true);
+    expect(service.getDocumentSecureMode()).toBe(false);
```
danlysiak, line 163: I've not looked into why this behaviour is different, should be reverted - just pushed with this since it's the only unit test which fails and wanted to verify integration tests on the pipeline.

danlysiak: Looking further, the default value is `false` so this is actually the correct expectation.

danlysiak: Looking further, the default value is `false` so this is actually the correct expectation. The fact it was passing before with `true` is probably a symptom of the issue that #3910 is looking to fix!

PR #4044, opened 2024-10-29
Force additional check on user presence in session

danlysiak:
### Jira link

### Checklist

### DIFF
--- a/api/workAllocation/index.ts
+++ b/api/workAllocation/index.ts
@@ -539,6 +539,7 @@ export async function getTaskNames(req: EnhancedRequest, res: Response): Promise
  */
 export async function getUsersByServiceName(req: EnhancedRequest, res: Response, next: NextFunction): Promise<void> {
   console.log(req?.session);
+  console.log(req?.session?.passport?.user?.userinfo);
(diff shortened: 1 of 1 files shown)


api/workAllocation/index.ts
```diff
@@ -538,6 +538,9 @@ export async function getTaskNames(req: EnhancedRequest, res: Response): Promise
  * getUsersByServiceName
  */
 export async function getUsersByServiceName(req: EnhancedRequest, res: Response, next: NextFunction): Promise<void> {
+  if (!req?.session?.passport?.user?.userinfo) {
```
danlysiak, line 541: From local/preview testing, `userinfo` is only present when the user is authenticated, so no need to verify any information inside it. 
I did consider checking for the presence of `uid` inside the `userinfo` to be a bit more specific, but it's sometimes `uid` and other times `id`, so we'd have to check for the presence of either - I didn't think worth the extra code/complication when it can be simplified down to this anyway, but happy to change if preferred.

andywilkinshmcts (approved):
Thanks Dan. Similar but slightly more specific to what I tried yesterday

PR #4087, opened 2024-10-30
DO NOT MERGE - Add tag

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2462

### Change description

Toolkit change to ensure dynamic radio list selection is correct

### Testing done

Checked the functionality of the dynamic radio list and that the value of the field does not change after fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "@angular/ssr": "^17.3.6",
     "@cucumber/cucumber": "^10.8.0",
     "@edium/fsm": "^2.1.2",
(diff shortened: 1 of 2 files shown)


Josh-HMCTS (approved):

andywilkinshmcts (changes requested):
Need to be very careful about updating toolkit at the moment. If you want this to be release soon then the toolkit version should be from the temporary toolkit release branch, in which case it should be 7.0.74-something

Josh-HMCTS (approved):

andywilkinshmcts (approved):

PR #4090, opened 2024-11-01
exui-2518-date field validation for month which should accept 1 to 12 value

RiteshHMCTS:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-2518

### Change description

date field validation for month which should accept 1 to 12 value


connorpgpmcelroy (approved):
LGTM

andywilkinshmcts (approved):
Weird. I've checked the ts file and it appears to be expecting 1-12, so this should be all that is needed

PR #4112, opened 2024-11-15
EXUI-2544 - initial check-in to enable WA tabs in the event of launch…

anthonydummer:
… darkly not returning values.

### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2544

### Change description

There have been instances in aat and demo where the calls to the LaunchDarkly feature service return no values, typically after an event that causes the reloading of the application.  It was decided that should this happen, with the exception of solicitors with role pui-case-manager, that in these rare instances, the displaying of the WA tasks and roles and access tabs would be preferential to them being missing.  Existing role access would then prevent users from fulfilling tasks they should not. 

### Testing done

This has been tested by manipulating the code to hit the new method and to enable WA tabs when no data is returned and to also not enable the tabs where there is no data and the role is pui-case-manager.

### Checklist


src/cases/containers/case-viewer-container/case-viewer-container.component.ts
```diff
@@ -129,6 +132,10 @@ export class CaseViewerContainerComponent implements OnInit {
   private enablePrependedTabs(features: WAFeatureConfig, userRoles: string[], supportedServices: string[], excludedRoles: string[]): boolean {
     const caseJurisdiction = this.caseDetails && this.caseDetails.case_type && this.caseDetails.case_type.jurisdiction ? this.caseDetails.case_type.jurisdiction.id : null;
     const caseType = this.caseDetails && this.caseDetails.case_type ? this.caseDetails.case_type.id : null;
+
+    if (this.launchDarklyError(features, supportedServices)) {
+      return !userRoles.includes('pui-case-manager');
```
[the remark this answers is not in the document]
anthonydummer: Done

andywilkinshmcts (dismissed):

RiteshHMCTS (dismissed):

StoneBenHMCTS (approved):

MunishSharmaHMCTS (approved):

StoneBenHMCTS (dismissed):

andywilkinshmcts (approved):

RiteshHMCTS (dismissed):

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
```diff
@@ -156,7 +170,10 @@ export class CaseViewerContainerComponent implements OnInit {
       // @ts-ignore
       map(([feature, userRoles, supportedServices, excludedRoles]: [WAFeatureConfig, string[]]) =>
         this.enablePrependedTabs(feature, userRoles, supportedServices, excludedRoles) ? this.prependedTabs : []),
-      catchError(() => this.prependedTabs$ = of([]))
+      catchError((error) => {
+        this.loggerService.log('case-viewer-container - Error in prependedCaseViewTabs', error);
+        return this.prependedTabs$ = of([]);
```
andywilkinshmcts, line 175: The return situation in this function is a bit weird - it's supposed to return the prepended tabs, but it doesn't  explicitly do so (unless there is an error), but it does update this.prependedTab$
I think you should sort this out - make it always return an Observable, but not update the member attribute, and make the calling function do that. Either that, or stop it returning the tabs and make sure that it is solely responsible for setting this.prependedTab$. At the moment you've got both

andywilkinshmcts (changes requested):

src/cases/containers/case-viewer-container/case-viewer-container.component.ts
anthonydummer: I made the change to set this.prependedTab$ in the method.  To be consistent I made a similar change to setAppendedCaseViewTabs().

andywilkinshmcts (approved):
That looks much more robust - thanks

StoneBenHMCTS (approved):

connorpgpmcelroy (approved):
LGTM

PR #4156, opened 2024-12-16
Tech/e UI 2135 role assignment valid

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2134
https://tools.hmcts.net/jira/browse/EXUI-2135

### Change description

Blacklist dangerous characters from user details and role assignments

### Testing done

Added plenty of unit tests to confirm blacklisting of dangerous characters does not cause problems with application


MunishSharmaHMCTS (dismissed):

api/utils.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { expect } from 'chai';
+import { allDoNotContainDangerousCharacters, hasNoDangerousCharacters, urlHasNoDangerousCharacters } from './utils';
+
+describe('api utils', () => {
+  describe('hasNoDangerousCharacters', () => {
```
andywilkinshmcts, line 5: please add tests for dangerous chars

api/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  for (const value of values) {
+    if (!hasNoDangerousCharacters(value)) {
+      // if one value contains dangerous characters, return false
+      return false;
+    }
+  }
+  return true;
+}
+
+export function hasNoDangerousCharacters(value: string): boolean {
```
andywilkinshmcts, line 11: I don't like this function name. Ideally it would be called hasUnacceptableChars and you would negate the return value. 
It is harder for humans to reason about a function called "notDoX" than a function called "DoX"
You also need to add in a comment why those chars are not accepted.

api/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  // the below only checks for the characters in the string
+  // return /^[^%<>^$//]+$/.test(value);
+  // substring approach below preferred
```
andywilkinshmcts, line 14: i don't understand "substring approach below preferred"

api/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  return /^(?!.*\/\*|.*\/\/|.*;|.*&|.*\?|.*<|.*\^|.*>).+$/.test(value);
```
andywilkinshmcts, line 15: I don't know why this is so complex. What is wrong with /[^%<>^$//]/.test(value)
This function appears to return true if any of the chars are present, so has the wrong logical name

andywilkinshmcts (changes requested):
Needs refactoring:
Functions that return a boolean should be called
hasXXX 
or 
isXXX 
and never hasNoXXX or isNoXXX

The regexs look to me as though they could be simpler, unless I'm missing some subtly

api/utils.spec.ts
connorpgpmcelroy: There are tests for dangerous characters already - do you mean more tests testing more of the dangerous characters?

api/utils.ts
connorpgpmcelroy: Ok can do that

api/utils.ts
connorpgpmcelroy: Will describe in response to comment below

api/utils.ts
connorpgpmcelroy: This regEx allows for removing dangerous substrings too e.g. '.*\/\/' removes '//' as being possible in any of the data and '.*\/\*' removes '/*'. The regEx you mention can only be used to remove dangerous characters, not substrings. In the list of dangerous possibilities you gave me some of them were '//' for example and removing '/' is not necessary. I can change it to the regEx above but we would not have that flexibility

andywilkinshmcts (approved):
This could do with a bit of refactoring but it should be OK for to test

MunishSharmaHMCTS (approved):

Josh-HMCTS (dismissed):

Josh-HMCTS (approved):

PR #4172, opened 2025-01-08
Update query-management-container.component.ts

olusegz07:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-2638

### Checklist


StoneBenHMCTS (approved):

StoneBenHMCTS (approved):

Josh-HMCTS (approved):

andywilkinshmcts (dismissed):
This is the simplest thing that could possibly work, though it has some restrictions. In particular it will only replace the first occurence of the placeholder in a url. That's probably OK, given that I can't think of a scenario where more than one place holder would be needed.

andywilkinshmcts (approved):

andywilkinshmcts (approved):
This works, but It would be nice if you could reuse the general interpolation code. The toolkit includes the [ccdLabelSubstitutor] directive that handles interpolation for general fields. It might be fairly easy to reuse this. I will raise a separate ticket for this.

RiteshHMCTS (approved):

Josh-HMCTS (approved):

PR #4220, opened 2025-01-31
EXUI-2483 - update coat of arms

jphcdcgi:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2483

### Change description

New coat of arms artwork.

### Checklist


andywilkinshmcts (changes requested):
It looks like the new image doesn't have a transparent background. I think this might be a problem. See if you can find a transparent version

jphcdcgi:
> It looks like the new image doesn't have a transparent background. I think this might be a problem. See if you can find a transparent version

We'll need a new link to a transparent PNG asset then. The Gov.uk deep link in the JIRA ticket doesn't have one.

PR #4221, opened 2025-02-03
Cot 887

sonny-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/COT-887 

### Change description

Default panel required based on a flag passed by the service else default to false


anthonydummer (approved):
LGTM

PR #4271, opened 2025-02-12
use new default screenflow if a paneldefault has been set

sonny-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/COT-942

only use the new screenflow in the event panelRequiredDefault is not undefined


anthonydummer (approved):
LGTM

PR #4363, opened 2025-03-17
exui-2889-v2

anthonydummer:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2889

### Change description

Added in a check on the validationErrors length for the instance where it has not been set to null and is an empty array. 

### Testing done

Button was seen to be enabled after a click before, but this change can be seen to be doing the additional check and correctly setting the cancelActioned flag, disabling the button after the first click. 

### Checklist

- [x] Does this PR introduce a breaking change - No


Josh-HMCTS (dismissed):

andywilkinshmcts (dismissed):
works, could be better

PR #4376, opened 2025-03-19
prevent an amend label where only change is case flag

anthonydummer:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2917

### Change description

A change we have made in EXUI-2321 has exposed a new bug for CR84.  When a new case flag has been created and accepted in the hearing edit, the hearing request is being allowed to progress.  However, if the only action is to accept the flag, and no changes to the request have been made, then this now gets a validation warning message to state no changes have been made.   This change is to remove the amended label for the section header, as this is misleading to the user. 

### Testing done

This has been tested with Doug Rice the QA and has been through a number of scenarios. 

### Checklist

### DIFF
--- a/src/hearings/utils/hearings.utils.ts
+++ b/src/hearings/utils/hearings.utils.ts
@@ -256,4 +256,29 @@ export class HearingsUtils {
       HearingsUtils.modifyHearingDetailsYear(input?.hearingDetails?.hearingWindow);
     }
   }
+
+  public static doArraysDiffer(array: string[], arrayToCompare: string[]): boolean {
+    return !_.isEqual(this.standardiseStringArray(array), this.standardiseStringArray(arrayToCompare));
+  }
+
+  private static standardiseStringArray(array: string[] | null | undefined): string[] | undefined {
+    return array && array.length > 0 ? [...array].sort() : undefined;
+  }
+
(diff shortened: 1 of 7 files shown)


StoneBenHMCTS (dismissed):

src/hearings/utils/hearings.utils.ts
```diff
@@ -256,4 +256,29 @@ export class HearingsUtils {
...
+
+  public static doArraysDiffer(array: string[], arrayToCompare: string[]): boolean {
+    return !_.isEqual(this.standardiseStringArray(array), this.standardiseStringArray(arrayToCompare));
+  }
+
+  private static standardiseStringArray(array: string[] | null | undefined): string[] | undefined {
+    return array && array.length > 0 ? [...array].sort() : undefined;
+  }
+
+  public static haveAdditionalFacilitiesChanged(
```
andywilkinshmcts, line 268: Please add a return type to this function, particularly as it's an odd one
should be :{Boolean, Boolean} by the looks of it

andywilkinshmcts (commented):
Looks like that should do what we talked about

andywilkinshmcts (dismissed):

Josh-HMCTS (dismissed):
LGTM

PR #4392, opened 2025-03-28
EXUI-2960 - hearings P2

andywilkinshmcts:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-2960

### Change description
Refactor ExUI to avoid using local storage to store the current case reference, and use the toolkit CaseNotifier to maintain a record of the current case. This should prevent problems caused by the session storage record of the current case getting out of sync with the user's display, for example by using the back button, or by rapidly switching tabs


andywilkinshmcts:
Closing as this has been replaced by https://github.com/hmcts/rpx-xui-webapp/pull/4413

PR #4480, opened 2025-05-28
Improve/ex UI 2654 judicial consistency

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-2654
https://tools.hmcts.net/jira/browse/EXUI-2907

### Change description

Refactor logic of isJudicial to rely on roleCategory and not user's roles (note that there are instances when user roles must be used - when looking for specific role or when listing a number of users)

### Testing done

Confirmed that this works correctly for judicial and non judicial users

  * [ ] Yes
  * [ x ] No


anthonydummer (commented):
Just a comment checking around the use of service related constant.

anthonydummer (approved):

andywilkinshmcts (changes requested):
There's a lot of valuable clean up here. but what I'm not sure about is why this PR still has logic that is specific to IAC judges. Also I don't really see where the code is that turns a user into a role category - did you move it to common lib?

connorpgpmcelroy:
> There's a lot of valuable clean up here. but what I'm not sure about is why this PR still has logic that is specific to IAC judges. Also I don't really see where the code is that turns a user into a role category - did you move it to common lib?

The logic specific to IAC judges (mentioned also by Anthony) is specific to testing, i.e. the named 'IAC_...' roles. The reason I did this was just in case the role name changed - it's used in quite a few tests so this may save someone time in future.

The role category is set within user details via the node layer within the refreshing of role assignments (set via role assignments or IDAM roles). The enum is set from common-lib just because there were a number of different enums that would have caused confusion and it's better to just use one across the application (apart from the node layer). I can move it if we want

andywilkinshmcts (approved):

andywilkinshmcts (approved):

PR #4521, opened 2025-06-30
EXUI-3247 spike to increase unit test coverage

chrisjones-hmcts:
### Jira link

See [EXUI-3247](https://tools.hmcts.net/jira/browse/EXUI-3247)

### Change description

Spike to see coverage increase in unit tests using AI - increased coverage in /api folder by around 15% in approx. 2.5 days of work. 

From:

image

to:

image


andywilkinshmcts (approved):
Very good start. Tests mostly make sense. Probably we should check for used code before doing any more

Josh-HMCTS (approved):

PR #4558, opened 2025-07-24
Add unsafe-inline removal from node-lib

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-3044

### Change description

Remove unsafe-inline to improve security

### Testing done

Confirmed no issues seen on the application

  * [ ] Yes
  * [ x ] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "@hmcts/nodejs-healthcheck": "1.8.5",
     "@hmcts/properties-volume": "^0.0.13",
     "@hmcts/rpx-xui-common-lib": "2.1.4",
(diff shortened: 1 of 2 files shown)


olusegz07:
If user is not navigating anywhere, is it better practice to use a < button > for actions, not an < a > tag.

connorpgpmcelroy:
> If user is not navigating anywhere, is it better practice to use a < button > for actions, not an < a > tag.

https://tools.hmcts.net/jira/browse/EXUI-3416

RiteshHMCTS (approved):

PR #4673, opened 2025-10-09
exui-3538-update ld flag value

RiteshHMCTS:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-3537

### Change description
Update LD flag value


olusegz07:
The value has been reassigned to WorkAllocationRelease2 in MC, as it appears that LD is using this value. There is no difference between this codebase and master

PR #4711, opened 2025-11-04
exui-3684-url link

RiteshHMCTS:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-3684

### Change description
url link issue


connorpgpmcelroy (changes requested):
Confirm why change needed and whether we need to consume multiple possible fields from API

PR #4750, opened 2025-11-19
EXUI-3066 - HALO-23226 dates times

chrisjones-hmcts:
### Jira link

See [EXUI-3066](https://tools.hmcts.net/jira/browse/EXUI-3066), [EXUI-3005](https://tools.hmcts.net/jira/browse/EXUI-3005), [EXUI-3003](https://tools.hmcts.net/jira/browse/EXUI-3003)

(corresponding toolkit PR: https://github.com/hmcts/ccd-case-ui-toolkit/pull/2046)

Changes across these two PRs:
Updated the datetime-picker component to use a dual control pattern, using a localDisplayControl to display local time to the user for interacting with, which is synced to the existing dateControl for UTC storage, using momentj.s for conversions.

Also replaced manual (native JS date) timezone offset calculations which could be vulnerable to daylight savings transitions with explicit moment.js based conversions.

  * [ ] Yes
  * [ x ] No


Josh-HMCTS (dismissed):
Looks fine to me, i like the approach of using one method (moment) to do dates instead of using a few different ways across everything.

RiteshHMCTS (dismissed):

RiteshHMCTS (dismissed):

PR #4827, opened 2025-12-23
exui-3815 min doc cve fix

balajisridharanhmcts:
### Jira link

See [EXUI-3815](https://tools.hmcts.net/jira/browse/EXUI-3815)

  * [ ] Yes
  * [ ] No

### Checklist


connorpgpmcelroy (dismissed):
Looks much better now

gregs-cgi:
It looks like `min-document` is only used in `global`, which is only used in `@edium/fsm`. I can't see `@edium/fsm` being used anywhere in the application (although I haven't checked super hard), is there a reason we're just adding a resolution instead of clearing down this whole dependency stack?

chrisjones-hmcts (dismissed):

gregs-cgi (changes requested):
It looks like min-document is only used in global, which is only used in @edium/fsm. I can't see @edium/fsm being used anywhere in the application (although I haven't checked super hard), is there a reason we're just adding a resolution instead of clearing down this whole dependency stack?

balajisridharanhmcts:
The @edium/fsm library is used in toolkit which when removed throws a build error. Also updating this library to latest (2.1.5) also doesn't remove the min-document vulnerability. So added "min-document": "^2.19.1" to resolution.

chrisjones-hmcts (approved):

PR #4932, opened 2026-02-10
EXUI-4114 MC leaking configuration data

chrisjones-hmcts:
### Jira link

See [EXUI-4114](https://tools.hmcts.net/jira/browse/EXUI-4114)

### Change description

pr to reduce exposure of unauthenticated configuration and remove redundant endpoints.

- removed unauthenticated endpoint `/external/config/check` (the broad app config snapshot route)
- removed open endpoint`/external/config/ui`, and kept `/external/configuration-ui` to pass config from the backend to ui - same pattern as in MO and AO
- moved `/api/healthCheck` behind auth middleware
- refactored Terms & Conditions feature-flag lookup to use preloaded runtime config instead of calling `/api/configuration`.
- added `termsAndConditionsEnabled` to the `/external/configuration-ui` payload.
- now getting sessionSecret from keyvault
- updated tests and test mocks to align with endpoint and config changes.

### Testing done

All unit and pipeline tests pass

  * [x] Yes
  * [ ] No

### DIFF
--- a/playwright-nightly.config.ts
+++ b/playwright-nightly.config.ts
@@ -61,9 +61,9 @@ module.exports = defineConfig({
   /* Retry on CI only */
   retries: 1, // Set the number of retries for all projects
 
-  timeout: 5 * 60 * 1000, // 5 minutes per test maximum as on first nightly run tests were taking too long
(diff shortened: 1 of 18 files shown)


connorpgpmcelroy (approved):
LGTM, this should work

PR #4943, opened 2026-02-13
Decentralised events

banderous:
### Change description

https://tools.hmcts.net/confluence/spaces/RRFM/pages/1923744323/Decentralised+professional+journeys

Suggest we merge this as an experimental feature for lower environments when happy

### Testing done

Tested locally with the Cftlib


src/cases/utils/decentralised-redirect.util.ts
```diff
@@ -0,0 +1,183 @@
+import { Params } from '@angular/router';
+import type { DecentralisedCaseTypeConfig, DecentralisedCaseTypeConfigMap } from '../../models/environmentConfig.model';
+
+const DECENTRALISED_EVENT_PREFIX = 'ext:';
+const TEMPLATE_PLACEHOLDER = '%s';
+const DECENTRALISED_NOC_PATH = '/noc';
+
+const isDecentralisedEvent = (eventId?: string): eventId is string => {
+  return !!eventId && eventId.startsWith(DECENTRALISED_EVENT_PREFIX);
```
jamescatto-cgi, line 8: Generally, encoding information in strings like this should be avoided because it makes it hard to understand what information is encoded, requires parsing and is prone to formatting errors. In this case, it also requires a hardcoded string to be configured. A better approach would be to use a separate parameter that indicates the intention, e.g. isDecentralised, which would also make this method unnecessary.

src/cases/utils/decentralised-redirect.util.ts
```diff
@@ -0,0 +1,183 @@
+import { Params } from '@angular/router';
```
jamescatto-cgi, line 1: Instead of having a `utility` file like this, that has all kinds of arbitrary objects in it, a more OO approach would group these functions into several classes using the Single Responsibility Principle. This would make this code easier to understand since the class/file structure itself gives clues about what functions are available and how they should be used.

jamescatto-cgi (commented):
I know this is just prototype code so feel free to ignore my comments - I'm just reading this for my own understanding.

src/cases/utils/decentralised-redirect.util.ts
banderous: A separate event specific marker would unfortunately require changing ccd definition store, data store and the ccd ui toolkit.

src/cases/utils/decentralised-redirect.util.ts
banderous: There aren't any objects in this file - these are pure functions. I don't see how introducing classes would improve it?

api/noc/index.ts
```diff
@@ -1,48 +1,129 @@
...
+function getCachedNoCCaseType(req: EnhancedRequest, caseId: unknown): string | null {
+  if (!req.session || caseId === undefined || caseId === null) {
+    return null;
+  }
+
+  const caseType = req.session[NOC_CASE_TYPE_SESSION_KEY]?.[String(caseId)];
+  return caseType || null;
+}
+
+function getConfiguredCaseType(caseTypeConfig: DecentralisedCaseTypeConfigMap, caseType: string): string | null {
```
jamescatto-cgi, line 114: This function is duplicated in `decentralised-redirect.util`, is there any way to remove this duplication?

api/noc/index.ts
```diff
@@ -1,48 +1,129 @@
...
+  const lowerCaseType = caseType.toLowerCase();
+  return (
+    Object.keys(caseTypeConfig)
+      .filter((configuredCaseType) => lowerCaseType.startsWith(configuredCaseType.toLowerCase()))
```
jamescatto-cgi, line 118: why not just match it entirely? Would getting the 1st value in this case always be correct?

api/noc/index.ts
```diff
@@ -1,48 +1,129 @@
...
+      .sort((first, second) => second.length - first.length)[0] || null
+  );
+}
+
+function resolveUrl(url: string, configuredCaseType: string, caseType: string): string {
```
jamescatto-cgi, line 123: This function is duplicated in `decentralised-redirect.util`, is there any way to remove this duplication?

api/noc/index.ts
```diff
@@ -1,48 +1,129 @@
...
-    const { status, data }: { status: number; data: NoCQuestions } = await handlePost(markupPath, body, req);
+    const baseUrl = getNoCBaseUrl(body?.case_id, req);
+    const { status, data }: { status: number; data: NoCQuestions } = await handlePost(`${baseUrl}/noc/noc-requests`, body, req);
     res.status(status).send(data);
   } catch (error) {
     next(generateErrorMessageWithCode(error));
   }
 }
+
+function getNoCBaseUrl(caseId: unknown, req: EnhancedRequest): string {
```
jamescatto-cgi, line 63: It might be better to move these functions into a separate class, which would simplify this index file and encapsulate the logic around the URL decision.

Josh-HMCTS (dismissed):
Looks good to me, if possible the comments left by James could be addressed in regards to duplicate code across multiple files just to ensure there is no drift between the files and makes future implementation easier if code is needed.

RiteshHMCTS (dismissed):

PR #4950, opened 2026-02-18
EXUI-4423 - fix(WA): harden downstream error handling and close WA service test gaps

andyg-cgi:
### Jira link

See [EXUI-4423](https://tools.hmcts.net/jira/browse/EXUI-4423)

### Change description

fix(wa): harden downstream error handling and close WA service test gap.

- centralize WA log+rethrow logic into shared work-allocation-error.utils
- refactor task/case/caseworker/supported-jurisdiction services to use shared helper
- preserve fallback UX by rethrowing errors after telemetry
- make searchTask spec assert currentTasks$ side-effect (avoid false-positive non-subscribed test)
- add missing getMyAccess success/failure coverage
- extend caseworker service failure-path tests for remaining endpoints
- keep unit suite green and prevent MonitoringService DI-related regressions in component specs

### Testing done

test run

  * [ ] Yes
  * [x] No

### Checklist


api/workAllocation/index.ts
```diff
@@ -581,33 +610,45 @@ export async function getTaskNames(req: EnhancedRequest, res: Response): Promise
...
+      return;
     }
+
+    const cachedUserData = await fetchUserData(req);
+    const cachedUsers = await fetchRoleAssignments(cachedUserData, req);
+    const refinedUsers = searchAndReturnRefinedUsers(services, term, cachedUsers);
+    res.status(200).send(refinedUsers);
   } catch (error) {
+    const term = req?.body?.term;
+    const services = Array.isArray(req?.body?.services) ? req.body.services : [];
```
connorpgpmcelroy: Disagree with this as of https://tools.hmcts.net/jira/browse/EXUI-2645 because the change for that is to ensure caseworker data from all services is not returned, services need to be specified by the angular layer. Therefore I'd say if there are no services in the call then we don't return the caseworkers. However since I don't know when I'll finish EXUI-2645 and this change isn't in affect for the calls currently i suggest we leave it as it is but add a TODO to not return the caseworkers if there are no services

src/work-allocation/services/work-allocation-task.service.ts
```diff
@@ -103,12 +162,31 @@ export class WorkAllocationTaskService {
...
+            error
+          )
+        )
+      );
     }
-    return of(null);
+    return of([]);
   }
 
   public getTaskTypeNamesFromService(): Observable<TaskNamesResponse[]> {
```
connorpgpmcelroy: This is dead code - can remove it

connorpgpmcelroy (changes requested):
2 minor changes requested. Also unsure about the logAndRethrow aspect of this change - namely the logging to appInsights. This could dump a lot more logs into appInsights, a lot of which will be unneccessary, although as it is proactive to put those logs in before we need them I don't disagree with the change

api/workAllocation/index.ts
andyg-cgi: Added an explicit empty-services guard that returns `[]`, with an EXUI-2645 TODO and route coverage. This keeps us from ever falling back to all caseworkers when services are omitted.

src/work-allocation/services/work-allocation-task.service.ts
andyg-cgi: Removed this unused method and the spec-only coverage for it.

api/workAllocation/caseWorkerUserDataCacheService.spec.ts
```diff
@@ -2,18 +2,25 @@ import * as chai from 'chai';
 import { expect } from 'chai';
```
Josh-HMCTS, line 2: I would be cautious around making changes in these caching files, also connor has some changes in progress around this area so may be better to remove these ones for time being and revisit at some point in the future

PR #5027, opened 2026-03-18
EXUI-4265-Integration tests - Update the case list positive tests

ShamrezDarw:
EXUI-4265-Integration tests - Update the case list positive tests


andyg-cgi (commented):
Left one inline comment on the page-4 pagination block. The main point is that the test currently proves the request contract for page 2 but not for page 4, and the suggestion swaps that second hop to the same request assertion helper so the coverage matches the intent of the change.

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
```diff
@@ -4,8 +4,16 @@ import { buildCaseListMock, buildCaseListJurisdictionsMock } from '../../mocks/c
...
+const PAGE_SIZE = 25;
+
 const caseListJurisdictionsMock = buildCaseListJurisdictionsMock();
 
+function getExpectedResultsSummary(totalResults: number, pageNumber: number, pageSize: number = PAGE_SIZE): string {
+  const startResult = (pageNumber - 1) * pageSize + 1;
+  const endResult = Math.min(totalResults, pageNumber * pageSize);
+  return `Showing ${startResult} to ${endResult} of ${totalResults} results`;
+}
+
```
danH-test: Test functions like this should live in the page object file.

andyg-cgi (dismissed):

andyg-cgi (commented):
A few follow-up comments from me on test strength and maintainability. I’ve called out the two gaps I mentioned earlier around false-positive coverage, and I’ve also left a few smaller notes on duplication and page-object boundaries with suggested snippets to make the intent clearer.

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
```diff
@@ -1,80 +1,483 @@
...
+    const visiblePageNumbers = await caseListPage.getVisiblePaginationPageNumbers();
+    const visibleIntermediatePageNumbers = visiblePageNumbers.slice(1, -1);
 
-    await test.step('Navigate to the search page', async () => {
-      await caseListPage.navigateTo();
-      await expect(caseListPage.exuiHeader.header).toBeVisible();
+    expect(visiblePageNumbers[0]).toBe(1);
+    expect(visiblePageNumbers.at(-1)).toBe(totalPages);
+    expect(visibleIntermediatePageNumbers).toEqual(
+      Array.from({ length: visibleIntermediatePageNumbers.length }, (_, index) => index + 2)
```
andyg-cgi: This part is doing more than the user contract really guarantees. The HMCTS toolkit paginator is rendered with a fixed window size and ellipsis support, so asserting that the visible intermediate numbers are a contiguous `2..N` run makes the test brittle to paginator internals rather than case-list behaviour.

I’d keep the warning/current page/request/rows assertions and drop the visible-number shape check, e.g.

```ts
const pageFourHundredSearchRequestPromise = waitForSearchCasesRequest(page, { page: 400 });
await caseListPage.clickPaginationPage(400);
const pageFourHundredSearchRequest = await pageFourHundredSearchRequestPromise;

expect(pageFourHundredSearchRequest).toEqual({
  ctid: 'xuiTestJurisdiction',
  useCase: 'WORKBASKET',
  view: 'WORKBASKET',
  page: '400',
});
await expectPaginationState(caseListPage, { currentPage: 400, previousVisible: true, nextVisible: false });
await expectCaseListSummary(caseListPage, totalResults, 400);
await expectCaseListRows(caseListPage, tableUtils, pageFourHundredMock);
```

That still proves the 10,000-result path without binding the spec to toolkit windowing details.

playwright_tests_new/E2E/page-objects/pages/exui/caseList.po.ts
```diff
@@ -79,6 +90,45 @@ export class CaseListPage extends Base {
...
+    await this.page.waitForFunction(
+      ({ selector, optionValueOrLabel: target }) => {
+        const element = document.querySelector(selector);
+        if (!(element instanceof HTMLSelectElement)) {
+          return false;
+        }
+
+        return Array.from(element.options).some((option) => option.value === target || option.label.trim() === target);
+      },
+      { selector: await select.evaluate((element) => `#${(element as HTMLSelectElement).id}`), optionValueOrLabel },
```
andyg-cgi: This helper is in the right page object, but I’d simplify the implementation. We already have the `Locator`, so round-tripping back through `document.querySelector('#id')` adds an unnecessary dependency on the element having a CSS-safe `id`.

A simpler pattern is to inspect the options straight off the locator and then select once you know whether you matched by value or label:

```ts
const matchingOption = await select.evaluate((element, target) => {
  const options = Array.from((element as HTMLSelectElement).options);
  const match = options.find((option) => option.value === target || option.label.trim() === target);
  return match ? { value: match.value, label: match.label.trim() } : null;
}, optionValueOrLabel);

if (!matchingOption) {
  throw new Error(`Dropdown option "${optionValueOrLabel}" was not available.`);
}

await select.selectOption(
  matchingOption.value === optionValueOrLabel
    ? matchingOption.value
    : { label: matchingOption.label }
);
```

That keeps the wait/inspection scoped to the locator itself and removes the global selector hop.

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
```diff
@@ -1,80 +1,483 @@
...
+  });
+
+  await page.route('**/data/internal/searchCases*', async (route) => {
+    if (options.searchResponseHandler) {
+      await options.searchResponseHandler(route);
+      return;
+    }
+
+    await route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify(options.searchResponse) });
   });
```
andyg-cgi: This is useful support code, but it has crossed the line where the repo guidance wants it out of the spec. The file now owns route setup, request waiting, table assertions, summary assertions, and pagination assertions, which makes the scenario body much harder to scan.

I’d extract the shared support into an integration helper and leave the spec with just the business flow, for example:

```ts
// integration/helpers/case-list.helper.ts
export async function setupCaseListMocks(page: Page, options: CaseListMockOptions): Promise<void> {
  // shared route wiring here
}

export async function expectCaseListState(args: {
  caseListPage: CaseListPage;
  tableUtils: TableUtils;
  mock: CaseListMockResponse;
  totalResults: number;
  pageNumber: number;
}): Promise<void> {
  await expectCaseListSummary(args.caseListPage, args.totalResults, args.pageNumber);
  await expectCaseListRows(args.caseListPage, args.tableUtils, args.mock);
}
```

That keeps the positive spec readable and gives the negative suite somewhere consistent to reuse the same setup when it grows.

andyg-cgi (commented):
Added three inline comments on the highest-signal issues I found:

- the brittle `>10000` pagination assertion that depends on toolkit windowing
- the dropdown helper in the page object, which can be simplified and scoped to the locator
- the amount of reusable route/assertion support left inside the spec instead of a shared integration helper

Each comment includes a concrete code sketch for the next edit.

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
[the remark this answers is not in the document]
ShamrezDarw: I’ve consolidated the positive spec into a single describe, extracted shared mock setup into caselist.helper.ts and to include the aggregated jurisdictions route, 
Added the explicit 10,000-results warning assertion, 
Updated the state-filter scenario to use the standard case list data shape rather than the special preset fixture.

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
ShamrezDarw: I removed the visible paginator-number shape assertions from the >10000 results test and kept the contract-level checks only

playwright_tests_new/E2E/page-objects/pages/exui/caseList.po.ts
ShamrezDarw: Updated selectDropdownOption to inspect options directly from the existing locator

playwright_tests_new/integration/test/caseList/caseList.positive.spec.ts
ShamrezDarw: I moved the shared case list setup/assertion support intointergration/helper/caseList.helper.ts

playwright_tests_new/E2E/page-objects/pages/exui/caseList.po.ts
```diff
@@ -24,8 +23,16 @@ export class CaseListPage extends Base {
...
+  readonly unselectableCasesInfoDetails = this.page.locator('#info-msg-unselected-case details');
+  readonly unselectableCasesInfoSummaryButton = this.page.locator('#info-msg-unselected-case summary');
+  readonly unselectableCasesInfoSummary = this.page.locator('#sp-msg-unselected-case-header');
+  readonly unselectableCasesInfoContent = this.page.locator('#sp-msg-unselected-case-content');
   readonly pagination = this.page.locator('.ngx-pagination');
-
+  readonly paginationNext = this.page.locator('.pagination-next');
+  readonly paginationPrevious = this.page.locator('.pagination-previous');
+  readonly paginationCurrentPage = this.pagination.locator('.current');
   // Some case list views use an id, others a data-test attribute for the summary
```
danH-test: playwright_tests_new/E2E/page-objects/components/
Should have these page objects already so we can reference them in our tests.

andyg-cgi (approved):

andyg-cgi (dismissed):

playwright_tests_new/E2E/page-objects/pages/exui/caseList.po.ts
ShamrezDarw: Changed to use the existing page objects

danH-test (approved):

RiteshHMCTS (approved):

PR #5065, opened 2026-04-01
EXUI-4143 E2E improvements add media viewer

ShamrezDarw:
A single positive-path E2E test that validates the Media Viewer flow end-to-end.
Test should cover the full happy path in the UI and be reliable against stable test data.
The test must use dedicated test data (case/document) to avoid flakiness and allow repeatable runs.


playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        'application/pdf',
+        MEDIA_VIEWER_FIXTURE_CONTENT,
+        createCasePage.fileUploadInput,
+        'latin1'
+      );
+      await createCasePage.clickContinueMultipleTimes(4);
+      await createCasePage.clickSubmitAndWait('after uploading media viewer document', {
+        maxAutoAdvanceAttempts: 3,
+      });
+      await expect(caseDetailsPage.caseAlertSuccessMessage).toBeVisible();
```
andyg-cgi: Nice to see this flow covered end-to-end. The risky bit is that this goes straight from submit to asserting the new row is there, while the existing V2 upload spec in this repo already treats the same update-case path as eventually consistent and wraps it in retry/polling. Without that, this is likely to fail on ordinary refresh lag rather than a real regression.

Suggested direction:
```ts
await retryOnTransientFailure(
  async () => {
    await caseDetailsPage.selectCaseDetailsTab('Tab 1');
    await caseDetailsPage.selectCaseAction(UPDATE_CASE_ACTION, {
      expectedLocator: createCasePage.fileUploadInput,
    });
    await createCasePage.uploadFile(
      documentFileName,
      'application/pdf',
      MEDIA_VIEWER_FIXTURE_CONTENT,
      createCasePage.fileUploadInput,
      'latin1'
    );
    await createCasePage.clickContinueMultipleTimes(4);
    await createCasePage.clickSubmitAndWait('after uploading media viewer document', {
      timeoutMs: 60_000,
      maxAutoAdvanceAttempts: 3,
    });
    await expect(caseDetailsPage.caseAlertSuccessMessage).toBeVisible({ timeout: 30_000 });
  },
  {
    maxAttempts: 2,
    onRetry: async () => {
      await caseDetailsPage.reopenCaseDetails(caseDetailsUrl);
    },
  }
);

await expect.poll(async () => {
  await caseDetailsPage.selectCaseDetailsTab('Tab 1').catch(() => undefined);
  const rowText = await caseDetailsPage.documentOneRow.innerText().catch(() => '');
  return rowText.includes(documentFileName);
}, { timeout: 45_000, intervals: [1_000, 2_000, 3_000] }).toBe(true);
```

That would make this much closer to the repo's existing upload stability pattern.

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  process.cwd(),
+  'playwright_tests_new/integration/testData/documents/case-file-view-document-delivery.pdf'
+);
+const MEDIA_VIEWER_FIXTURE_CONTENT = readFileSync(MEDIA_VIEWER_FIXTURE_PATH, 'latin1');
+
+function findMediaViewerPage(pages: Page[]): Page | undefined {
+  return pages.find((candidate) => MEDIA_VIEWER_ROUTE_PATTERN.test(candidate.url()));
+}
+
+test.describe('Media Viewer happy path', { tag: ['@e2e', '@e2e-media-viewer'] }, () => {
```
andyg-cgi, line 22: Small CI/operability gotcha here: `@e2e-media-viewer` is new, but `playwright_tests_new/E2E/tag-filter.json` is not updated in this PR. The repo validates include/exclude overrides against that allow-list, so targeted Jenkins reruns or temporary exclusions for this suite will fail even though the test itself exists.

Suggested companion change:
```json
{
  "excludedTags": [],
  "availableTags": [
    "@e2e",
    "@e2e-case-flags",
    "@e2e-create-case",
    "@e2e-document-upload",
    "@e2e-manage-tasks",
    "@e2e-media-viewer",
    "@e2e-search-case",
    "@e2e-update-case"
  ]
}
```

Worth fixing in the same PR so this suite can actually be isolated through the supported tag-filter path.

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        .poll(() => {
+          const mediaPage = findMediaViewerPage(page.context().pages());
+          return mediaPage?.url() ?? '';
+        })
+        .toMatch(MEDIA_VIEWER_ROUTE_PATTERN);
+
+      const mediaPage = findMediaViewerPage(page.context().pages()) ?? page;
+      await mediaPage.waitForLoadState('domcontentloaded').catch(() => undefined);
+
+      await expect(mediaPage).toHaveURL(MEDIA_VIEWER_ROUTE_PATTERN);
```
andyg-cgi: The shell assertions are a useful smoke check, but they do not prove the correct document binary was actually requested and loaded. In this codebase the wrapper title comes from session/local storage before the fetch completes, so this can still pass even if the viewer opens with stale or broken document data.

Suggested direction:
```ts
const binaryResponses: string[] = [];
page.on('response', (response) => {
  if (response.request().method() === 'GET' && /\/documents(v2)?\/.+\/binary/.test(response.url())) {
    binaryResponses.push(response.url());
  }
});

await caseDetailsPage.openDocumentOne();

await expect.poll(() => binaryResponses.length).toBeGreaterThan(0);
await expect
  .poll(() => binaryResponses.at(-1) ?? '')
  .toContain('/documents');

const mediaPage = findMediaViewerPage(page.context().pages()) ?? page;
await expect(mediaPage.locator('exui-media-viewer')).toBeVisible();
await expect(mediaPage.locator('#mvToolbarMain')).toBeVisible();
```

If you can capture the exact binary URL or another document-specific loaded signal, that would make the test prove the real happy path instead of only the wrapper route.

andyg-cgi (commented):
Added three inline review comments with concrete fix directions and code snippets:
- stabilize the V2 upload flow using the repo’s existing retry/polling pattern
- register the new `@e2e-media-viewer` tag in the E2E tag-filter allow-list
- assert the media viewer binary request/load path, not just the wrapper shell

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+const UPDATE_CASE_ACTION = 'Update case';
+const MEDIA_VIEWER_FIXTURE_PATH = path.resolve(
+  process.cwd(),
+  'playwright_tests_new/integration/testData/documents/case-file-view-document-delivery.pdf'
+);
+const MEDIA_VIEWER_FIXTURE_CONTENT = readFileSync(MEDIA_VIEWER_FIXTURE_PATH, 'latin1');
+
+function findMediaViewerPage(pages: Page[]): Page | undefined {
+  return pages.find((candidate) => MEDIA_VIEWER_ROUTE_PATTERN.test(candidate.url()));
+}
```
danH-test: Ideally we should move this test function into a helper or PO file.

ShamrezDarw: Moved this into the Case Details PO

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: added @e2e-media-viewer to the E2E tag filter config.

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: Added a binary assertion so the test now proves the document fetch as well as the the viewer shell.

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: Updated to use the existing retry pattern used by the upload flow test

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  'playwright_tests_new/integration/testData/documents/case-file-view-document-delivery.pdf'
+);
+const MEDIA_VIEWER_FIXTURE_CONTENT = readFileSync(MEDIA_VIEWER_FIXTURE_PATH, 'latin1');
+
+test.describe('Media Viewer happy path', { tag: ['@e2e', '@e2e-media-viewer'] }, () => {
+  test.beforeAll(async ({ browserName: _browserName }, _testInfo) => {
+    await ensureSession('SOLICITOR');
+  });
+
+  test('opens auploaded document in the Media Viewer end-to-end', async ({ page, createCasePage, caseDetailsPage }, testInfo) => {
```
danH-test: minor typo and capitalisation

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        await mediaPage.waitForLoadState('domcontentloaded').catch(() => undefined);
+
+        await expect.poll(() => binaryResponses.length).toBeGreaterThan(0);
+        await expect.poll(() => binaryResponses.at(-1) ?? '').toMatch(/\/documents(?:v2)?\/[^/]+\/binary$/);
+
+        await expect(mediaPage).toHaveURL(MEDIA_VIEWER_ROUTE_PATTERN);
+        await expect.poll(async () => mediaPage.title()).toContain(`${documentFileName} - View Document`);
+        await expect(mediaPage.locator('exui-media-viewer')).toBeVisible();
+        await expect(mediaPage.locator('#mvToolbarMain')).toBeVisible();
+        await expect(mediaPage.locator('#viewerContainer')).toBeVisible();
```
danH-test: Can be in the PO File

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+      };
+
+      page.context().on('response', onResponse);
+      try {
+        const mediaPage = await caseDetailsPage.openDocumentOneInMediaViewer();
+        await mediaPage.waitForLoadState('domcontentloaded').catch(() => undefined);
+
+        await expect.poll(() => binaryResponses.length).toBeGreaterThan(0);
+        await expect.poll(() => binaryResponses.at(-1) ?? '').toMatch(/\/documents(?:v2)?\/[^/]+\/binary$/);
```
danH-test: could possibly be a const as we're using it more than once

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    await test.step('Authenticate and upload a PDF for this test run', async () => {
+      await ensureAuthenticatedPage(page, 'SOLICITOR', { waitForSelector: 'exui-header' });
+    });
+
+    await test.step('Create a case for this test run', async () => {
+      await createDivorceCase(createCasePage, JURISDICTION, CASE_TYPE, caseMarker);
+      caseDetailsUrl = await caseDetailsPage.getCurrentPageUrl();
+    });
+
+    await test.step('Upload a  document through the update flow', async () => {
```
danH-test: spacing in the step description

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: Moved to caseFileView PO

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: changed to const

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: removed the spacing

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: changed

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+const UPDATE_CASE_ACTION = 'Update case';
+const MEDIA_VIEWER_FIXTURE_PATH = path.resolve(
+  process.cwd(),
+  'playwright_tests_new/integration/testData/documents/case-file-view-document-delivery.pdf'
+);
+const MEDIA_VIEWER_FIXTURE_CONTENT = readFileSync(MEDIA_VIEWER_FIXTURE_PATH, 'latin1');
+
+test.describe('Media Viewer happy path', { tag: ['@e2e', '@e2e-media-viewer'] }, () => {
+  test.beforeAll(async ({ browserName: _browserName }, _testInfo) => {
+    await ensureSession('SOLICITOR');
```
danH-test: Should be apply session function here

playwright_tests_new/E2E/page-objects/pages/exui/caseFileView.po.ts
```diff
@@ -128,6 +131,16 @@ export class CaseFileViewPage extends Base {
...
+  public async expectStandaloneMediaViewerLoaded(expectedTitleText?: string): Promise<void> {
+    if (expectedTitleText) {
+      await expect.poll(async () => this.page.title()).toContain(`${expectedTitleText} - View Document`);
+    }
+
+    await expect(this.standaloneMediaViewerContainer).toBeVisible();
+    await expect(this.standaloneMediaViewerToolbar).toBeVisible();
+    await expect(this.standaloneMediaViewPanel).toBeVisible();
+  }
+
```
danH-test: No expects outside of the test.spec. 
Can we have this in the test instead as this is only used/required once?

ShamrezDarw: moved to test.spec

playwright_tests_new/E2E/test/mediaViewer/mediaViewerHappyPath.spec.ts
ShamrezDarw: session function added

andyg-cgi (dismissed):

danH-test (dismissed):

RiteshHMCTS (approved):

kasi-subramaniam (approved):

balajisridharanhmcts (commented):
LGTM

PR #5075, opened 2026-04-09
test: stabilise PRL hearings E2E setup

kasi-subramaniam:
### Jira link

See [EXUI-3622](https://tools.hmcts.net/jira/browse/EXUI-3622)

### Change description ###

This PR adds stable Playwright coverage for the PRL hearings happy-path journey and hardens the supporting setup so the test can run against a real, eligible PRL case instead of relying on global search finding a reusable case by chance.

The main changes are:

- Adds the PRL user hearings E2E journey for submitting a new hearing.
- Adds dynamic PRL hearings case setup for CI and local runs when `PRL_HEARINGS_CASE_SETUP=true`.
- Resolves the previous challenged-access failure mode where global search found PRL cases, but the hearing manager could not open a usable case-details tab list.
- Updates the setup contract to use the correct RPX/PRL Key Vault-backed environment variables.
- Ensures PRL setup mints a fresh `ccd_data` S2S token from `S2S_URL` / `PRL_HEARINGS_S2S_URL`, instead of trusting a generic app `S2S_TOKEN`.
- Adds unit coverage around the PRL setup env parsing, required configuration checks, and S2S-token behaviour.
- Wires the PRL-only Jenkins secrets into the Playwright E2E shell environment using `withAzureKeyvault`, scoped only to the E2E step when PRL setup is enabled.
- Adds the missing `idam-web-url` to the PRL Jenkins secret mapping so Jenkins provides `IDAM_WEB_URL` to the setup fixture.

### Why this helps

For developers:

- The PRL hearings journey can be reproduced locally with a deterministic setup path instead of depending on whichever old PRL case happens to appear in global search.
- Setup failures now fail early with a clear missing-variable message, making broken local or CI configuration easier to diagnose.
- The fixture accepts the actual RPX/PRL env aliases used by Key Vault and local `.env` files, reducing one-off local workarounds.

For testers and SDETs:

- The test now creates or resolves a case that is intended to be usable by the hearing manager, reducing challenged-access noise.
- The supporting setup has direct API/unit coverage, so future env-contract regressions are caught faster than waiting for a full browser journey to fail.
- The PRL setup is explicitly opt-in via `PRL_HEARINGS_CASE_SETUP=true`, so unrelated Playwright E2E, integration, API, accessibility, and nightly paths are not forced to carry PRL setup dependencies.

For the project:

- This removes a known fragile area in the hearings regression pack and gives the PR pipeline a clearer signal: if PRL setup credentials or services are missing, the failure points directly at the setup contract.
- The Jenkins changes are scoped to the PRL E2E execution path and avoid broad pipeline behaviour changes.
- The branch keeps product/runtime code untouched; changes are limited to Playwright setup/tests and the targeted Jenkins secret wiring needed to run them.

### Testing done

Automated:

- [x] `git diff --check` - passed
- [x] `PLAYWRIGHT_SKIP_INSTALL=true yarn lint` - passed with `0 errors`; existing repo warning baseline remains
- [x] `PLAYWRIGHT_SKIP_INSTALL=true PLAYWRIGHT_REPORTERS=list yarn test:api:pw playwright_tests_new/api/unit/prl-hearings-case-setup.unit.api.ts` - `11 passed`
- [x] `PRL_HEARINGS_CASE_SETUP=true PLAYWRIGHT_SKIP_INSTALL=true PLAYWRIGHT_REPORTERS=list yarn playwright test -c playwright.e2e.config.ts playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts --retries=0 --workers=1` - targeted PRL hearings E2E passed locally
- [x] `PRL_HEARINGS_CASE_SETUP=true PLAYWRIGHT_SKIP_INSTALL=true PLAYWRIGHT_REPORTERS=list yarn test:playwrightE2E` - full Playwright E2E passed locally: `17 passed`, `1 skipped`
- [x] `PLAYWRIGHT_SKIP_INSTALL=true PLAYWRIGHT_REPORTERS=list yarn test:playwright:integration` - full Playwright integration passed locally: `254 passed`, `7 skipped`

Manual / CI checks:

- [x] Verified all referenced PRL setup secrets exist in `rpx-aat` without printing secret values.
- [x] Confirmed `rpx-preview` does not exist; preview E2E should use `rpx-aat`, matching the pipeline's preview-to-AAT vault convention.
- [ ] Jenkins PR rerun still required after the latest Jenkins secret mapping commit.

Evidence:

- Local E2E Odhín report: `functional-output/tests/playwright-e2e/odhin-report/xui-playwright-e2e.html`
- Local integration Odhín report: `functional-output/tests/playwright-integration/odhin-report/xui-playwright-integration.html`

  * [ ] Yes
  * [x] No

No new dependency or CVE suppression is introduced by this PR.

### Checklist

- [x] README and other documentation has been updated / added if needed
- [x] tests have been updated / new tests have been added if needed
- [x] Does this PR introduce a breaking change: No


playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      await hearingsCYAPage.verifyHearingSummarySection(page, 'Additional instructions', [
+        { key: 'Enter any additional instructions for the hearing', value: 'Additional instructions for E2E Playwright test' },
+      ]);
+
+      console.log('~~~~~~~~~~ CYA page Hearings ... assertions done  ');
+    });
+
+    // Submit Hearings / Check Confirmation / Return to Hearing Summary page.
+    await test.step('Submit Hearing Request And Do Checks on backend calls ', async () => {
+      try {
```
andyg-cgi, line 210: This `try/catch` is the first thing I’d tighten up. If the submit call never happens or returns anything other than `201`, the test should fail immediately. Logging and carrying on makes the journey look healthier than it is.

Suggested shape:

```ts
const submitResponsePromise = page.waitForResponse(
  (response) => response.url().includes('/api/hearings/submitHearingRequest') && response.request().method() === 'POST'
);

await hearingsJourneyPage.submitRequestButton.click();
const submitResponse = await submitResponsePromise;

expect(submitResponse.status(), 'submitHearingRequest should return 201').toBe(201);
```

If we want extra diagnostics, we can attach the response body, but I’d still let the assertion fail rather than swallowing it.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      await hearingsCYAPage.verifyHearingSummarySection(page, 'Additional facilities', [
+        { key: 'Will additional security be required?', value: 'Yes' },
+        { key: 'Select any additional facilities required', value: additionalFacilitiesValue as AdditionalFacility[] }, //pad like 'xxx,yyy'
+      ]);
+
+      await hearingsCYAPage.verifyHearingSummarySection(page, 'Stage', [
+        { key: 'What stage is this hearing at?', value: 'Allocation' },
+      ]);
+
+      // await hearingsCYAPage.verifyHearingSummarySection(page, 'Participant attendance', [
```
andyg-cgi: I wouldn’t leave these CYA checks commented out in the PR. The happy path is doing a lot of form work before this page, so the CYA page is where we prove those choices survived the journey.

I’d either make these active now, or remove the commented block and leave the genuinely out-of-scope variations for the follow-up ticket. For this PR, I’d expect at least participant attendance and venue to be asserted:

```ts
await hearingsCYAPage.verifyHearingSummarySection(page, 'Participant attendance', [
  { key: 'Will this be a paper hearing?', value: 'No' },
  { key: 'What will be the methods of attendance for this hearing?', value: ['Video', 'Telephone'] },
  { key: 'How will each participant attend the hearing?', value: ['Video', 'Telephone'] },
  { key: 'How many people will attend the hearing in person?', value: '2' },
]);

await hearingsCYAPage.verifyHearingSummarySection(page, 'Hearing venue', [
  { key: 'What are the hearing venue details?', value: /Southampton/i },
]);
```

The exact venue value can be adjusted to what the autocomplete reliably selects, but the section should not be untested.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        console.error('Failure seen on  Hearings Submit CYA Page :', error);
+      }
+      // if all good then reached here on confirmation page./;
+      await hearingsJourneyPage.checkHearingConfirmationPage(page);
+
+      await hearingsJourneyPage.clickLinkToViewHearings(page);
+
+      await expect(page).toHaveURL(/\/cases\/case-details\/.*#Hearings$/);
+      await expect(hearingsTabPage.currentAndUpcomingHeading('Current and upcoming')).toBeVisible();
+      hearingId = await hearingsJourneyPage.getMostRecentHearingId();
```
andyg-cgi: At this point I’d pin the assertions to the row we just created, not just “some first row exists”. Otherwise an old hearing on the case can make the test pass while the new request is not reflected correctly.

Something along these lines would make the end of the happy path much stronger:

```ts
const hearingId = await hearingsJourneyPage.getMostRecentHearingId();
const hearingRow = hearingsTabPage.hearingRow(hearingId, 'view-details');

await expect(hearingRow).toContainText('Awaiting hearing details');
await expect(hearingsTabPage.viewDetailsButton(hearingId)).toBeVisible();

await hearingsTabPage.openViewDetails(hearingId);
await expect(page).toHaveURL(/\/hearings\/request\/hearing-view-summary$/);
await hearingViewSummaryPage.waitForReady();
await expect(hearingViewSummaryPage.sectionHeading('Hearing details')).toBeVisible();
```

If the current toggle state expects edit summary instead, use the existing `openViewOrEdit` / `hearingViewEditSummaryPage.waitForReady()` path and assert `/hearing-view-edit-summary$/`. The important bit is that the test proves the route and summary contract, not only that a numeric ID exists.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,290 @@
...
+    // hearingMethod
+    const hearingMethodAndLocatorMap: Record<HearingMethod, Locator> = {
+      'In Person': this.hearingInPerson,
+      Telephone: this.hearingOnTelephone,
+      Video: this.hearingViaVideo,
+    };
+
+    for (const method of hearingMethod ?? []) {
+      const locator = hearingMethodAndLocatorMap[method];
+      await locator.click({ force: true });
```
andyg-cgi: I’d avoid `force: true` here. If Playwright cannot click the checkbox normally, that is exactly the sort of actionability problem this E2E should expose rather than bypass.

Suggested shape:

```ts
for (const method of hearingMethod) {
  const locator = hearingMethodAndLocatorMap[method];
  await expect(locator, `${method} hearing method should be selectable`).toBeVisible();
  await locator.check();
}
```

If these are not native checkboxes in practice, use the label/role locator that a real user would activate, but keep it as a normal user-level action.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,290 @@
...
+      'Not in Attendence': this.hearingInPerson,
+    };
+    const party1Select = this.firstParty;
+    party1Select.selectOption('Video');
+
+    const party2Select = this.secondParty;
+    party2Select.selectOption('Telephone');
+
+    // how many will Attend.
+    this.numberAttendingHearing.fill('2');
```
andyg-cgi: These three Playwright actions need `await`. Without it, the test can hit Continue before the participant selections and attendee count have actually been applied, which will be painful to debug in CI.

Suggested fix:

```ts
await this.firstParty.selectOption(model.get('hearingAttendence', 'attendHearingHow')?.[0] ?? 'Video');
await this.secondParty.selectOption(model.get('hearingAttendence', 'attendHearingHow')?.[1] ?? 'Telephone');
await this.numberAttendingHearing.fill(model.get('hearingAttendence', 'numberOfPeopleAttendingHearing') ?? '2');
```

Even better, keep the values entirely model-driven so the CYA assertions and form input cannot drift away from each other.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  async setHearingVenue(model: HearingJourneyModel): Promise<void> {
+    const hearingVenue = model.get('hearingVenue', 'name') as string;
+
+    await this.hearingVenue.pressSequentially(hearingVenue, { delay: 1000 });
```
andyg-cgi: The one-second delay makes this slower and still does not really prove the venue list is ready. I’d wait for the actual option/result we want, then select it.

For example:

```ts
await this.hearingVenue.fill(hearingVenue);
const option = this.page.getByRole('option', { name: /Southampton/i }).first();
await expect(option).toBeVisible();
await option.click();
await this.addLocationsButton.click();
```

The matcher can be whatever venue is reliable for PRL, but the wait should be state-based rather than time-based.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,290 @@
...
+        await this.specificDateNo.click();
+        break;
+      case 'Yes':
+        await this.specificDateYes.click();
+        break;
+      default:
+        await this.specificDateRange.click();
+        break;
+    }
+    this.setHearingPriority(hearingPriority);
```
andyg-cgi: This should be awaited, and the helper below needs to honour the requested priority. At the moment the helper always checks Standard, so an Urgent variant would silently exercise the wrong path.

Suggested fix:

```ts
await this.setHearingPriority(hearingPriority);
```

and:

```ts
async setHearingPriority(priority: string): Promise<void> {
  const option = priority === 'Urgent' ? this.hearingPriorityUrgent : this.hearingPriorityStandard;
  await option.check();
}
```

andyg-cgi (commented):
One general thing I could not put inline because the file is not in the diff: please register the new E2E tag in `playwright_tests_new/E2E/tag-filter.json`, otherwise the suite is awkward to target through the repo’s normal tag entrypoint.

```json
{
  "availableTags": [
    "@e2e-prl-hearings"
  ]
}
```

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
kasi-subramaniam: try catch wrapped around correctly with error handling

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
kasi-subramaniam: this has been addressed now on Participant Attendence section

kasi-subramaniam: Hearing Venue location section is also being verified in the CYA page

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
kasi-subramaniam: hearingId from API response is compared with the hearingId of the mostRecently created hearing ( as taken from the UI ) . Hearing Row checks for Status and View Details link is also being done.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: 1.  await added on line 153 to 159
2.  Form values are now being retrieved from the HearingModel

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: this has been addressed now.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,294 @@
...
+    // hearingMethod
+    const hearingMethodAndLocatorMap: Record<HearingMethod, Locator> = {
+      'In Person': this.hearingInPerson,
+      Telephone: this.hearingOnTelephone,
+      Video: this.hearingViaVideo,
+    };
+
+    for (const method of hearingMethod ?? []) {
+      const locator = hearingMethodAndLocatorMap[method];
+      await locator.click({ force: true });
```
andyg-cgi: I still would not keep `force: true` here. If these controls are not actionable in the real UI, that is exactly the thing I want the E2E to expose rather than step around.

I would keep this on a normal user path and let Playwright fail on the real contract:

```ts
for (const method of hearingMethod ?? []) {
  const locator = hearingMethodAndLocatorMap[method];

  await expect(locator, `${method} hearing method should be visible`).toBeVisible();
  await expect(locator, `${method} hearing method should be enabled`).toBeEnabled();
  await locator.check();
}
```

If these are not native checkbox inputs in practice, I would still switch to the user-facing control for the same option, for example a label or role-based locator, rather than falling back to `force: true`. The important bit is that we prove a real user-level interaction works.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,294 @@
...
+    await party2Select.selectOption(howIsHearingAttended[1] as string);
+
+    const noOfPeopleAttending = model.get('hearingAttendence', 'numberOfPeopleAttendingHearing') as string;
+    await this.numberAttendingHearing.fill(noOfPeopleAttending);
+  }
+
+  async setHearingVenue(model: HearingJourneyModel): Promise<void> {
+    const hearingVenue = model.get('hearingVenue', 'name') as string[];
+
+    await this.hearingVenue.pressSequentially(hearingVenue[0], { delay: 1000 });
```
andyg-cgi: This is the one I still would not sign off on as-is. The one-second delay plus `ArrowDown` / `Enter` is still a timing workaround, not proof that the venue autocomplete has actually resolved the option we want.

I would rather drive the real option and fail if it never appears:

```ts
async setHearingVenue(model: HearingJourneyModel): Promise<void> {
  const [requestedVenue] = model.get('hearingVenue', 'name') as string[];

  await this.hearingVenue.fill(requestedVenue);

  const venueOption = this.page
    .getByRole('option')
    .filter({ hasText: new RegExp(requestedVenue, 'i') })
    .first();

  await expect(
    venueOption,
    `Venue option for "${requestedVenue}" should appear in the autocomplete`
  ).toBeVisible();

  await venueOption.click();
  await this.addLocationsButton.click();
  await expect(this.removeLocationLink(requestedVenue)).toBeVisible();
}
```

If the display label is longer than the typed prefix, I would still wait on the rendered listbox row and click that specific option. I would not keep the sleep-based path here.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      console.log('~~~~~~~~~~ CYA page Hearings ... assertions done  ');
+    });
+
+    // Submit Hearings / Check Confirmation / Return to Hearing Summary page.
+    await test.step('Submit Hearing Request And Do Checks on backend calls ', async () => {
+      let submitHearingResponse: any;
+      let apiResponseHearingId: string;
+
+      try {
```
andyg-cgi, line 210: This is better than the earlier version because it rethrows, but I still think the happy path should stay straight-line here and fail directly on the main submit contract. The broad `try/catch` makes the core assertion harder to read and easier to grow into a catch-all over time.

I would keep the main path as the thing we are actually asserting, then move any optional response-body diagnostics into a small helper if we still need them:

```ts
const submitResponsePromise = page.waitForResponse(
  (response) =>
    response.url().includes('/api/hearings/submitHearingRequest') &&
    response.request().method() === 'POST'
);

await hearingsJourneyPage.submitRequestButton.click();
const submitHearingResponse = await submitResponsePromise;

expect(submitHearingResponse.status(), 'submitHearingRequest should return 201').toBe(201);

const submitBody = await submitHearingResponse.json();
const apiResponseHearingId = submitBody.hearingRequestID;
```

If you want the extra failure detail as well, I would keep that in a dedicated helper that is only called after a failed status assertion, rather than wrapping the whole success path in one big `try/catch`.

andyg-cgi (commented):
I’ve left a few targeted comments where I think the PR needs tightening before it is comfortable to approve. The theme is not that the journey is the wrong thing to test; it is that a couple of assertions and interaction helpers currently make the test less trustworthy than it should be for a long E2E path.

Jenkinsfile_CNP
andyg-cgi, line 1: I’d keep this PR focused on the hearings E2E work and move the Playwright install behaviour change out of this diff. Removing `PLAYWRIGHT_SKIP_INSTALL=true` changes how broad CI jobs install browsers for integration, preview E2E, and AAT E2E runs. That may be the right fix for a pipeline problem, but it needs its own evidence and review because it affects more than this one hearing journey.

Suggested fix: revert the Jenkinsfile changes here. If the browser install change is needed, raise a small CI PR that includes the failing job evidence, the exact command that needs browser installation, and confirmation that the extra install cost is acceptable for the affected jobs.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsCYA.po.ts
```diff
@@ -0,0 +1,279 @@
...
+    // There could be pre-selected venues as well which are part of the CaseCreation flow which we may or may not control
+    // Hence it will be apt to do a 'contains' check here vis-a-vis a 'exact' match.
+
+    const expectedRows: Array<{ key: string; value: string | string[]; match?: 'exact' | 'contains' }> = [
+      { key: 'What are the hearing venue details?', value: hearingVenue, match: 'contains' },
+    ];
+    await this.verifyHearingSummarySection(this.page, 'Hearing Venue', expectedRows);
+  }
+
+  public async verifyHearingLocation(page: Page, hearingJourneyModel) {
```
andyg-cgi: This method name says it verifies the hearing location, but the implementation below is another participant-attendance assertion block. That means the spec can select or render the wrong venue and still pass the CYA “location” step.

Suggested fix: either call the existing venue-specific helper from the spec, or change this method to assert the `Hearing Venue` section and the `What are the hearing venue details?` row. A shape like this would make the contract explicit:

```ts
public async verifyHearingVenueSection(hearingJourneyModel: HearingJourneyModel): Promise<void> {
  const hearingVenue = hearingJourneyModel.get('hearingVenue', 'name');
  if (!Array.isArray(hearingVenue)) {
    throw new Error(`hearingVenue must be an array, got ${typeof hearingVenue}`);
  }

  await this.verifyHearingSummarySection(this.page, 'Hearing Venue', [
    { key: 'What are the hearing venue details?', value: hearingVenue, match: 'contains' },
  ]);
}
```

Then the spec should call `verifyHearingVenueSection(...)`, not a participant-attendance duplicate.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
andyg-cgi: I’d avoid `force: true` here. It gets the test past Playwright actionability, but it also hides the exact browser state we need this E2E to prove. If the checkbox is covered, disabled, detached, or not actually user-clickable, this test should tell us that rather than forcing the click through.

Suggested fix: make the locator target the actual input or label that a user can interact with, wait for it to be visible/enabled, and use a normal `check()`/`click()`:

```ts
for (const method of hearingMethod) {
  const locator = hearingMethodAndLocatorMap[method];
  await expect(locator, `${method} hearing method should be selectable`).toBeVisible();
  await locator.check();
}
```

If the current locator is not checkable without force, that is useful feedback that we should tighten the locator rather than bypass actionability.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
andyg-cgi: This venue selection still feels too timing-dependent for a critical happy-path journey. Typing with a one-second delay and then pressing `ArrowDown` / `Enter` proves the current list ordering and timing, not that the expected venue option resolved.

Suggested fix: fill the search term, wait for the specific option, click that option, and then assert the selected venue appears before continuing:

```ts
async setHearingVenue(model: HearingJourneyModel): Promise<void> {
  const [requestedVenue] = model.get('hearingVenue', 'name') as string[];

  await this.hearingVenue.fill(requestedVenue);

  const venueOption = this.page
    .getByRole('option')
    .filter({ hasText: new RegExp(requestedVenue, 'i') })
    .first();

  await expect(venueOption, `Venue option for "${requestedVenue}" should appear`).toBeVisible();
  await venueOption.click();
  await this.addLocationsButton.click();
  await expect(this.removeLocationLink(requestedVenue)).toBeVisible();
}
```

If the rendered option text is longer than the prefix, that is fine; we should still click the specific rendered option instead of relying on keyboard position.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  test('Submit a new Hearing - Happy Path journey ', async ({
+    caseDetailsPage,
+    hearingsTabPage,
+    hearingViewSummaryPage,
+    hearingsJourneyPage,
+    hearingsCYAPage,
+    page,
+  }) => {
+    hearingRouteConfig.caseReference = '1775575928056201'; // hardcoding for now.
```
andyg-cgi: This hard-coded case reference makes the E2E dependent on one live mutable case staying in exactly the right state. For a long journey that creates a real hearing request, that is a brittle dependency and it makes failures harder to diagnose.

Suggested fix: move the hearing scenario data into a dedicated hearings test-data module and resolve the case through the repo’s setup conventions, or at least centralise the fixed seeded case with a clear name and readiness assertion. For example:

```ts
const hearingScenario = prlHearingHappyPathScenario();
const caseReference = await resolvePrlHearingCase(page, hearingScenario);

await page.goto(caseDetailsUrl(
  hearingScenario.jurisdictionId,
  hearingScenario.caseTypeId,
  caseReference
));
```

That keeps the spec focused on the journey and gives us one place to document what state the case must have.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+      const hearingRow = hearingsTabPage.hearingRow(hearingId, 'view-or-edit');
+
+      await expect(hearingRow).toContainText('WAITING TO BE LISTED');
+      await expect(hearingsTabPage.viewOrEditButton(hearingId)).toBeVisible();
+    });
+
+    console.log('~~~~~~~~~~  Created Hearing with ID  === ' + hearingId);
+  });
+
+  function setUpHearingJourneyData() {
```
andyg-cgi: The amount of scenario data below is now big enough that I’d move it out of the spec. The spec should read like the hearing journey and its business assertions; the data factory belongs in a named support module.

Suggested fix: create something like `playwright_tests_new/E2E/testData/hearings/hearingJourneyScenarios.ts` or `playwright_tests_new/E2E/mocks/hearings/hearingJourney.mock.ts` and return a typed scenario object:

```ts
export const prlHearingHappyPath = {
  facilities: { additionalSecurity: 'Yes', additionalFacilities: ['Custody Cell', 'Laptop', 'Projector', 'Witness Room'] },
  stage: 'Allocation',
  attendance: { paperHearing: 'No', hearingMethod: ['Video', 'Telephone'], attendHearingHow: ['Video', 'Telephone'], peopleInPerson: '2' },
  venue: ['Southamp'],
  details: { hearingInWelsh: 'No', specificJudge: 'No', judgeType: ['Deputy High Court Judge', 'Deputy Circuit Judge'] },
  duration: { days: 0, hours: 1, minutes: 30 },
} satisfies HearingJourneyScenario;
```

Then the spec can initialise the model from that fixture, which makes future negative or Welsh variants much easier to add without duplicating the journey body.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsCYA.po.ts
```diff
@@ -0,0 +1,279 @@
...
+    //*[contains(@class, 'govuk-summary-list__row')]//*[contains(@class, 'govuk-summary-list__key') and contains(text(), 'What will be the methods of attendance for this hearing?')]
+    //await expect(this.getRowChangeLink(sectionTitle, key)).toBeVisible();
+  }
+
+  // Verify that the Section exists
+  async verifySectionVisible(sectionTitle: string) {
+    await expect(this.getSection(sectionTitle)).toBeVisible();
+  }
+
+  async verifyHearingSummarySection(
```
andyg-cgi: This helper is doing a lot of scenario verdict work inside the page object. I can see why it is useful, but in this repo the cleaner split is: page object exposes CYA rows/sections, and the spec owns the business expectations. That makes it easier for reviewers to see exactly what the journey proves.

Suggested fix: keep small locators/helpers here, for example `summarySection(title)`, `summaryRow(section, key)`, or `summaryValue(section, key)`, and move the expected-row arrays plus `expect(...)` calls back to the spec or a dedicated assertion module. A lighter page-object API might be:

```ts
summaryValue(sectionTitle: string, key: string): Locator {
  return this.summaryRow(sectionTitle, key).locator('.govuk-summary-list__value');
}
```

Then the spec can say `await expect(hearingsCYAPage.summaryValue('Stage', 'What stage is this hearing at?')).toHaveText(/Allocation/)`, which keeps the verdict visible.

andyg-cgi (commented):
Thanks for pushing the hearings journey forward. I’ve left the comments in a direct but fix-focused shape: the main concern is not the value of the E2E, it is that a few pieces currently make the test harder to trust than it needs to be.

The must-fix areas for me are: keep the Jenkins/browser-install change out of this test PR unless it has its own CI evidence, make the CYA venue assertion actually verify the venue section, remove the forced/timing-based UI interactions, and move the heavier scenario data out of the spec. Once those are tightened, this will be much easier to sign off as a stable SDET contribution rather than a one-off happy path.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
kasi-subramaniam: expect(submitHearingResponse.status(), 'submitHearingRequest should return 201').toBe(201);
has been included as a checkpoint.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: lines 137 to 164 has been added to address the above issue around the hearingVenue selection. force:true has been removed.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: Code around the line 155 of the hearingJourney.po.ts  has been amended now.

playwright_tests_new/E2E/utils/test-setup/prlHearingsCaseSetup.ts
```diff
@@ -0,0 +1,503 @@
...
+      config,
+      courtAdminUserId,
+      courtAdminToken,
+      serviceToken,
+      caseReference,
+      ISSUE_AND_SEND_TO_LOCAL_COURT_EVENT_ID,
+      buildIssueAndSendToLocalCourtEventData(config)
+    );
+    await getCaseInfo(apiContext, config, courtAdminToken, serviceToken, caseReference);
+    logger.info('Retry reset navigation failed before create-case beforeEach retry', { error });
```
andyg-cgi: This one is a real build issue, not just a style nit. This line looks like it was copied from the create-case retry block, but here it is in the happy path, so both `logger` and `error` are out of scope. That is enough to break the PR build.

I would keep the fix very small and just remove the line:

```ts
await getCaseInfo(apiContext, config, courtAdminToken, serviceToken, caseReference);

return caseReference;
```

No need to add a new logger here unless we actually want a successful setup log. If we do want that, use the existing pattern in this file instead:

```ts
createLogger().info(`PRL hearings setup - created usable case ${caseReference}.`);
return caseReference;
```

But honestly I would just delete it. The `extractCaseReference` helper already logs the created case reference, so adding another success log is not buying us much.

kasi-subramaniam: createLogger().info() call removed. 
    logger.info('Retry reset navigation failed before create-case beforeEach retry', { error }) removed.

playwright_tests_new/E2E/utils/test-setup/hearingsCaseResolver.ts
```diff
@@ -0,0 +1,170 @@
...
+    const accessHint =
+      probeStatus === 'challenged-access'
+        ? ' It opened the challenged-access screen, so check the created case court location against the hearing manager work area.'
+        : '';
+    throw new Error(
+      `PRL hearings setup created case ${createdCaseReference}, but it did not open a usable case-details tab list for ${route.jurisdictionId}/${route.caseTypeId}.${accessHint} The resolver validates access in the signed-in hearing manager session, so check the setup user's location and the hearing manager access model before falling back to shared cases.`
+    );
+  }
+
+  const candidateCaseReferences = await resolveCandidateCaseReferences(page, route);
```
andyg-cgi: I think this is the one bit I would still tighten before merge. Because this journey submits a hearing request, falling back to global search means we can end up mutating whichever existing PRL case happens to be usable for the hearing manager. That makes the test data non-deterministic and a bit risky for shared environments.

I’d make this journey fail fast unless it has either an explicit case reference or the dynamic PRL setup switched on. Something like:

```ts
const createdCaseReference = await createPrlHearingsCaseIfEnabled();

if (createdCaseReference) {
  await openCaseDetailsProbe(page, route, createdCaseReference);
  const probeStatus = await getCaseDetailsProbeStatus(page);

  if (probeStatus === 'usable') {
    return createdCaseReference;
  }

  const accessHint =
    probeStatus === 'challenged-access'
      ? ' It opened the challenged-access screen, so check the created case court location against the hearing manager work area.'
      : '';

  throw new Error(
    `PRL hearings setup created a case, but it did not open a usable case-details tab list for ${route.jurisdictionId}/${route.caseTypeId}.${accessHint}`
  );
}

throw new Error(
  'PRL hearings journey needs PRL_HEARINGS_CASE_REFERENCE or PRL_HEARINGS_CASE_SETUP=true. Do not fall back to global search for a journey that submits a hearing request.'
);
```

So basically I’d remove the global-search fallback from this mutating journey. It is fine for read-only discovery, but for a submit flow it feels too easy to damage shared seeded data by accident.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    const hearingVenue = model.get('hearingVenue', 'name') as string[];
+    const venueSearchTerm = hearingVenue?.[0];
+
+    if (!venueSearchTerm) {
+      throw new Error(`Expected a hearing venue search term, got: ${JSON.stringify(hearingVenue)}`);
+    }
+
+    await this.hearingVenue.fill(venueSearchTerm);
+    const venueOption = this.page.getByRole('option').filter({ hasText: venueSearchTerm }).first();
```
andyg-cgi, line 138: Small one, but I think this would make the test much more meaningful. At the moment we search for `Southampton`, click the first matching option, then assert the CYA page contains whatever text came back from that option. That proves the UI kept its own selected value, but it does not prove we selected the intended venue.

I’d put the expected venue in the scenario and select/assert that exact label:

```ts
hearingVenue: {
  searchTerm: 'Southampton',
  expectedVenue: 'Southampton Combined Court Centre'
}
```

```ts
const venueSearchTerm = model.get('hearingVenue', 'searchTerm') as string;
const expectedVenue = model.get('hearingVenue', 'expectedVenue') as string;

await this.hearingVenue.fill(venueSearchTerm);

const venueOption = this.page.getByRole('option', { name: expectedVenue });
await venueOption.waitFor({ state: 'visible', timeout: 30_000 });
await venueOption.click();
await this.addLocationsButton.click();

return expectedVenue;
```

Then the spec can keep asserting `selectedVenue`, but it will now be the expected business value rather than whichever autocomplete option happened to be first.

playwright_tests_new/integration/test/restrictedAccess/restrictedCaseAccess.negative.spec.ts
```diff
@@ -29,6 +31,23 @@ const globalSearchResultsHandler = createGlobalSearchResultsRouteHandler({
...
+  await expect(page.getByRole('heading', { level: 2, name: 'Users with access' })).toBeVisible();
+
+  const table = caseDetailsPage.exuiBodyComponent.table;
+  await expect(table.locator('thead th, thead td')).toHaveCount(3);
+  await expect(table.locator('thead th, thead td').nth(0)).toHaveText('User');
+  await expect(table.locator('thead th, thead td').nth(1)).toHaveText('Case role');
+  await expect(table.locator('thead th, thead td').nth(2)).toHaveText('Email address');
+  await expect(table.locator('tbody tr')).toHaveCount(0);
+}
+
```
danH-test, line 41: Two things here

1. We should keep test functions in another file away from specs so they can be reused
2. We shouldn't have expects in a test function, which is essentially what's happening here, sometimes it's ok to duplicate code if we're performing checks like these some of the locator strategies could be tightened here, but otherwise we should keep these steps as is.

danH-test (dismissed):

playwright_tests_new/integration/test/restrictedAccess/restrictedCaseAccess.negative.spec.ts
kasi-subramaniam: This has now been addressed.  a separate helper has been created to move the expect( ) checks - restrictedCaseAccessContainer.helper.ts has got all these checks being done now , hence keeping the function expectRestrictedAccessShellWithoutRows() succint.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: this has now been fixed. code ensures that Southampton Combined courts is part of the checks for hearing venue in the CYA Page. This ensures what has been chosen as part of the hearing flow is confirmed verified at the CYA Page

playwright_tests_new/E2E/utils/test-setup/hearingsCaseResolver.ts
kasi-subramaniam: This comment is addressed . The global-search fallback has been removed from  from this mutating journey.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+      await expect(
+        hearingsCYAPage.rowValue('Additional instructions', 'Enter any additional instructions for the hearing')
+      ).toHaveText(scenario.additionalInstructions);
+    });
+
+    await test.step('Submit Hearing Request And Do Checks on backend calls ', async () => {
+      let submitHearingResponse: any;
+      let apiResponseHearingId: string;
+
+      try {
```
andyg-cgi, line 210: Thanks for all the work that has gone into this. This is in a much stronger place now, especially with the CYA checks and the API response being tied back to the UI row.

One last thing I would still simplify here before closing this thread: I think the submit path would be easier to review if it stayed as a straight-line assertion. The current try/catch does rethrow, so it is not hiding the failure now, but it makes the main contract a bit harder to follow.

I think this shape would keep the important checks clearer:

```ts
const submitResponsePromise = page.waitForResponse(
  (response) =>
    response.url().includes('/api/hearings/submitHearingRequest') &&
    response.request().method() === 'POST'
);

await hearingsJourneyPage.submitRequestButton.click();

const submitHearingResponse = await submitResponsePromise;

expect(submitHearingResponse.status(), 'submitHearingRequest should return 201').toBe(201);

const submitBody = await submitHearingResponse.json();
const apiResponseHearingId = submitBody.hearingRequestID;

expect(apiResponseHearingId, 'submitHearingRequest should return a hearingRequestID').toBeTruthy();
```

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -0,0 +1,228 @@
...
+  async setHearingVenue(model: HearingJourneyModel): Promise<string> {
+    const hearingVenue = model.get('hearingVenue', 'name') as string[];
+    const venueSearchTerm = hearingVenue?.[0];
+
+    if (!venueSearchTerm) {
+      throw new Error(`Expected a hearing venue search term, got: ${JSON.stringify(hearingVenue)}`);
+    }
+
+    await this.hearingVenue.pressSequentially(venueSearchTerm);
+    const venueOption = this.page.getByRole('option').filter({ hasText: venueSearchTerm }).first();
```
andyg-cgi, line 138: Thanks for tightening this up. Waiting for the autocomplete option is a good improvement over the earlier timing-based approach.

The only remaining thing I would change is to make the expected venue explicit in the scenario. At the moment we search for `Southampton`, click the first option that contains that text, and then assert the selected value later. That does prove the selected value carries through, but I think naming the expected option directly would make this more robust and easier to maintain.

Something like this would make the intended venue explicit:

```ts
// hearing-model.ts
export interface HearingVenueData {
  searchTerm?: string;
  expectedVenue?: string;
}
```

```ts
// hearingJourneyScenarios.ts
hearingVenue: {
  searchTerm: 'Southampton',
  expectedVenue: 'Southampton Combined Court Centre',
},
```

```ts
// hearingJourneyScenarios.ts satisfies block
hearingVenue: {
  searchTerm: string;
  expectedVenue: string;
};
```

```ts
// createHearingJourneyModel()
model.set('hearingVenue', 'searchTerm', scenario.hearingVenue.searchTerm);
model.set('hearingVenue', 'expectedVenue', scenario.hearingVenue.expectedVenue);
```

```ts
// hearingsJourney.po.ts
async setHearingVenue(model: HearingJourneyModel): Promise<string> {
  const venueSearchTerm = model.get('hearingVenue', 'searchTerm');
  const expectedVenue = model.get('hearingVenue', 'expectedVenue');

  if (!venueSearchTerm || !expectedVenue) {
    throw new Error(
      `Expected hearing venue search term and expected venue, got: ${JSON.stringify({
        venueSearchTerm,
        expectedVenue,
      })}`
    );
  }

  await this.hearingVenue.fill(venueSearchTerm);

  const venueOption = this.page.getByRole('option').filter({ hasText: expectedVenue }).first();
  await venueOption.waitFor({ state: 'visible', timeout: 30_000 });

  await venueOption.click();
  await this.addLocationsButton.click();

  return expectedVenue;
}
```

andyg-cgi (commented):
Focused follow-up on the remaining review threads. These are intended to help get the PR to a merge-ready state.

playwright_tests_new/integration/test/restrictedAccess/restrictedCaseAccess.negative.spec.ts
```diff
@@ -29,6 +33,11 @@ const globalSearchResultsHandler = createGlobalSearchResultsRouteHandler({
 const RESTRICTED_ACCESS_MESSAGE = 'This case is restricted. The details of the users with access are provided below.';
 const RESTRICTED_ACCESS_FAILURE_STATUSES = [400, 403, 500] as const;
 
+async function expectRestrictedAccessShellWithoutRows(page: Page, caseDetailsPage: CaseDetailsPage): Promise<void> {
+  await expect(page).toHaveURL(new RegExp(`/cases/restricted-case-access/${VALID_SEARCH_CASE_REFERENCE}`));
+  await restrictedCaseAccessContainerHelper(page, caseDetailsPage);
```
andyg-cgi, line 39: Thanks for picking this one up. I can see what you were aiming for by moving the repeated block out of the spec.

I think we can keep that reuse while still making the assertions easy to see in the test. One option would be for the helper to return the relevant locators and expected value, then the spec keeps the `expect(...)` calls.

That would look something like:

```ts
// restrictedCaseAccessContainer.helper.ts
export function restrictedCaseAccessContainer(page: Page, caseDetailsPage: CaseDetailsPage) {
  const table = caseDetailsPage.exuiBodyComponent.table;

  return {
    restrictedMessage: page.getByText(RESTRICTED_ACCESS_MESSAGE),
    mainHeading: caseDetailsPage.exuiBodyComponent.mainHeading,
    container: caseDetailsPage.restrictedAccessContainer,
    usersWithAccessHeading: page.getByRole('heading', { level: 2, name: 'Users with access' }),
    tableHeaders: table.locator('thead th, thead td'),
    tableRows: table.locator('tbody tr'),
    expectedCaseReference: formatCaseNumberWithDashes(VALID_SEARCH_CASE_REFERENCE),
  };
}
```

```ts
// in the spec step
const restrictedAccess = restrictedCaseAccessContainer(page, caseDetailsPage);

await expect(page).toHaveURL(new RegExp(`/cases/restricted-case-access/${VALID_SEARCH_CASE_REFERENCE}`));
await expect(restrictedAccess.restrictedMessage).toBeVisible();
expect(await restrictedAccess.mainHeading.textContent()).toContain(restrictedAccess.expectedCaseReference);
await expect(restrictedAccess.container).toBeVisible();
await expect(restrictedAccess.usersWithAccessHeading).toBeVisible();

await expect(restrictedAccess.tableHeaders).toHaveCount(3);
await expect(restrictedAccess.tableHeaders.nth(0)).toHaveText('User');
await expect(restrictedAccess.tableHeaders.nth(1)).toHaveText('Case role');
await expect(restrictedAccess.tableHeaders.nth(2)).toHaveText('Email address');
await expect(restrictedAccess.tableRows).toHaveCount(0);
```

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: The search term for hearing venue  'scenario.hearingVenue.searchTerm'  is set as 'Southampton' here  hearingJourneyScenarios.ts line 36.
In the journey , the above value is populated here hearingsJourney.po.ts in the  setHearingVenue() method . i.e: The First value that is matched against the term 'Southampton' is set as the hearingVenue. 

On CYA Page, we verify that the above selected value is what is seen prior to actual submission of hearing. This proves that selected venue search is what is seen on the CYA page as well.

Jenkinsfile_CNP
kasi-subramaniam: only the PRL specific Azure keys that are required for the test to be run are mentioned here . Other references have been removed.

playwright_tests_new/integration/test/restrictedAccess/restrictedCaseAccess.negative.spec.ts
kasi-subramaniam: The above suggested changes have been made in the code

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
kasi-subramaniam: suggested changes have been made around the SearchTerm for the hearing location Venue field - and this has been passing on successful runs on both local and on pipeline.

playwright_tests_new/E2E/test/hearings/hearings-journey.spec.ts
kasi-subramaniam: this change has been made as suggested above on the test.step('Submit Hearing Request And Do Checks on backend calls');

andyg-cgi (approved):

ShamrezDarw (approved):

connorpgpmcelroy (approved):
LGTM

PR #5128, opened 2026-05-05
EXUI-4225 Add DynaTrace Agentless Tag

nathan-clark:
https://tools.hmcts.net/jira/browse/EXUI-4225

Adds optional support for a Dynatrace JavaScript tag for 'Agentless' monitoring.

The script URL is read from the DYNATRACE_CDN environment variable through existing config mapping and is gated by FEATURE_DYNATRACE_ENABLED. If the feature flag is enabled and the variable is set, the app injects the CDN URL into the rendered HTML and the existing nonce-protected bootstrap script appends the external Dynatrace script. If the flag is disabled or the variable is not set, no Dynatrace script is added.

Flux PR to manage prod config: https://github.com/hmcts/cnp-flux-config/pull/45509

NOTE:

`DYNATRACE_CDN` and `FEATURE_DYNATRACE_ENABLED` have been removed from `charts/xui-webapp/values.preview.template.yaml` since this has completed QA.


charts/xui-webapp/values.aat.template.yaml
```diff
@@ -19,3 +19,4 @@ nodejs:
     HEALTH_EM_DOCASSEMBLY_API: http://dg-docassembly-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/health
     HEALTH_S2S_SERVICE: http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/health
     HEALTH_TERMS_AND_CONDITIONS_API: http://xui-terms-and-conditions-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/health
+    DYNATRACE_CDN: ''
```
connorpgpmcelroy: Is this line necessary?

config/default.json
```diff
@@ -112,6 +112,7 @@
     }
   },
   "protocol": "https",
+  "dynatraceCdn": "",
```
connorpgpmcelroy, line 115: Similar to above, is this line necessary?

connorpgpmcelroy (changes requested):
Questions regarding empty string usage in config - is it required? Can the dynatrace not be presumed empty?

charts/xui-webapp/values.aat.template.yaml
chrisjones-hmcts: From my understanding, these settings inherit from and override the values defined in `values.yaml`. If we don't explicitly provide an override here, the production Dynatrace URL will be inherited and end up being configured in AAT.

config/default.json
chrisjones-hmcts: This will be the default in all environments; prod will be managed by Flux

charts/xui-webapp/values.preview.template.yaml
```diff
@@ -31,6 +31,7 @@ nodejs:
     SERVICES_IDAM_SERVICE_OVERRIDE: false
     SERVICES_IDAM_ISS_URL: https://forgerock-am.service.core-compute-idam-aat2.internal:8443/openam/oauth2/realms/root/realms/hmcts
     FEATURE_OIDC_ENABLED: true
+    DYNATRACE_CDN: https://js-cdn.dynatrace.com/jstag/17177a07246/bf24054dsx/55f97384adadaf88_complete.js
```
chrisjones-hmcts: This is temporary just for QA purposes - will be removed before moving to Ready to Deploy

charts/xui-webapp/values.aat.template.yaml
chrisjones-hmcts: Removed this

connorpgpmcelroy (dismissed):

olusegz07 (dismissed):

chrisjones-hmcts:
At Usman Ali's request, I removed the DYNATRACE_CDN value (and set FEATURE_DYNATRACE_ENABLED  to false) from non prod environments since testing has been completed

olusegz07 (approved):

connorpgpmcelroy (approved):

PR #5134, opened 2026-05-06
EXUI-4542-E2E-dataLossCivilCasesCreateCaseFlags

ShamrezDarw:
Validate that executing the Create Case Flag event does not modify or remove existing case data when the case is already in Mediation state.


playwright_tests_new/E2E/page-objects/pages/exui/caseFlag.po.ts
danH-test, line 1: We should avoid expects in the PO files and keep them in the test.
Timeouts : if we're using the same everywhere, it should ideally be set elsewhere (it's worth checking the default before setting this.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      await expect(page).toHaveURL(/\/cases\/case-details\//);
+      await page.goto(`/cases/case-details/CIVIL/CIVIL/${caseNumber}#Summary`);
+      await caseDetailsPage.exuiSpinnerComponent.wait();
+      await expect(page).toHaveURL(new RegExp(`/cases/case-details/CIVIL/CIVIL/${caseNumber}`));
+    });
+
+    await test.step('Create party-level Other case flag for Claimant 1', async () => {
+      await caseFlagPage.openCreateCaseFlagEvent(baselineCaseDetails, caseNumber);
+      await caseFlagPage.completePartyOtherCaseFlagForClaimant1(TEST_FLAG_COMMENT);
+      await page.getByRole('button', { name: /^Submit$/i }).click();
```
danH-test: Should this be a Page object ?

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+const DEFAULT_BUSINESS_PROCESS_WAIT_INTERVAL_MS = 3_000;
+const DEFAULT_CCD_EVENT_TRIGGER_WAIT_TIMEOUT_MS = 420_000;
+const DEFAULT_CCD_EVENT_TRIGGER_WAIT_INTERVAL_MS = 3_000;
+const DEFAULT_CIVIL_ROLE_ASSIGNMENT_WAIT_TIMEOUT_MS = 60_000;
+const DEFAULT_CIVIL_S2S_MICROSERVICE = 'civil_service';
+const DEFAULT_CIVIL_CLAIMANT_EMAIL = 'civilmoneyclaimsdemo@gmail.com';
+const DEFAULT_CIVIL_DEFENDANT_EMAIL = 'cuiuseraat@gmail.com';
+const DEFAULT_CIVIL_CASE_FLAGS_SETUP_ALIAS = 'CIVIL_SOLICITOR';
+const DEFAULT_CIVIL_COURT_STAFF_ALIAS = 'CIVIL_COURT_STAFF';
+const DEFAULT_CIVIL_COURT_STAFF_EMAIL = 'hearing_center_admin_reg1@justice.gov.uk';
```
danH-test: is this a valid user ? should we be setting this in code ?

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
danH-test, line 1: Consider splitting this file as it contains a number of different helpers, suggest moving out the API data setup helpers elsewhere as a starter.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+    event_token: eventToken,
+    ignore_warning: false,
+  };
+
+  const submitPath = `data/cases/${encodeURIComponent(options.caseNumber)}/events`;
+  const submitResponse = await options.page.request.post(submitPath, {
+    data: eventBody,
+    failOnStatusCode: false,
+    headers: CCD_CREATE_EVENT_HEADERS,
+    timeout: 60_000,
```
danH-test: Hard set timeout here that could be universally set by a const/env elsewhere maybe?
This occurs a few times in the file so mentioning here once for brevity.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+  if (configured) {
+    return trimTrailingSlash(configured);
+  }
+
+  const ccdDataStoreUrl = firstNonEmpty(process.env.CCD_DATA_STORE_URL, process.env.SERVICES_CCD_DATA_STORE_API);
+  if (ccdDataStoreUrl?.includes('civil-cui-data-store-staging.aat.platform.hmcts.net')) {
+    return 'http://civil-service-aat.service.core-compute-aat.internal';
+  }
+  if (ccdDataStoreUrl?.includes('ccd-data-store-api-aat.service.core-compute-aat.internal')) {
+    return 'http://civil-service-aat.service.core-compute-aat.internal';
```
danH-test: hardcoded urls here should be avoided and in general, can this be inferred or set once and referenced when needed?

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+        partyEmail: user.email,
+        partyPhone: '07446777177',
+        primaryAddress: {
+          AddressLine1: '123',
+          AddressLine2: 'Fake Street',
+          AddressLine3: '',
+          PostCode: 'S12eu',
+          PostTown: 'sheffield',
+        },
+        type: 'INDIVIDUAL',
```
danH-test: Can we use faker here for some of this data unless it needs to be hardcoded?

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+            },
+          ],
+        },
+        applicant1RejectedRepaymentReason: 'reasons',
+      },
+      applicant1LiPResponseCarm: {
+        isMediationContactNameCorrect: 'No',
+        alternativeMediationContactPerson: 'new contact person',
+        isMediationEmailCorrect: 'No',
+        alternativeMediationEmail: 'anotherem@ail.com',
```
danH-test: if we do need to hardcode email address can we use @example.com

danH-test:
Ideally this test should only be running on the nightly e2e test run as it looks like it contains a lot of steps that might hold up PR runs.

andyg-cgi (commented):
The main issues are around credential/artifact safety, proving the exact Mediation/Create Case Flag contract, and making the new support code small enough to review and maintain.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+    return error instanceof Error ? error : new Error(String(error));
+  }
+}
+
+async function getIdamAccessToken(page: Page, config: CivilApiConfig, user: CivilApiUser): Promise<string> {
+  let lastError: Error | undefined;
+  for (let attempt = 1; attempt <= 5; attempt++) {
+    try {
+      const response = await page.request.post(
+        `${config.idamApiUrl}/loginUser?username=${encodeURIComponent(user.email)}&password=${encodeURIComponent(user.password)}`,
```
andyg-cgi: Can we avoid putting the IDAM username and password in the URL here? Playwright traces, proxy logs and failed request diagnostics can all capture the full URL, so this can leak reusable credentials into CI artifacts.

A safer shape would be to send credentials in the body, or better, reuse the existing token helper if the repo already has one:

```ts
const response = await page.request.post(`${config.idamApiUrl}/loginUser`, {
  failOnStatusCode: false,
  form: {
    username: user.email,
    password: user.password,
  },
  timeout: 60_000,
});
```

If this endpoint genuinely only supports query params, I think we need a documented exception and redaction around request logging before this goes in.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      const normalisedUpdatedCaseDetails = normaliseCaseDataForDataLossComparison(updatedCaseDetails, {
+        ignoredFlagComment: TEST_FLAG_COMMENT,
+      });
+
+      expect.soft(resolveCaseNumberFromPayload(updatedCaseDetails)).toBe(caseNumber);
+      expect
+        .soft(resolveCcdCaseStateId(updatedCaseDetails), 'Post-flag case state should remain Mediation')
+        .toBe(MEDIATION_STATE);
+
+      await testInfo.attach('civil case before create flag.json', {
```
andyg-cgi: I do not think we should attach the full CCD case payloads to the Playwright result. The generated Civil case includes contact details and case data, and these attachments can be retained in HTML/Odhín/JUnit/report storage.

Can we attach a redacted comparison report instead, and keep raw JSON behind an explicit local-only debug flag if it is really needed?

```ts
const report = buildDataLossComparisonReport({
  baselineCaseDetails: redactCivilCaseDetails(baselineCaseDetails),
  updatedCaseDetails: redactCivilCaseDetails(updatedCaseDetails),
  caseNumber: redactCaseReference(caseNumber),
  normalisedBaseline: normalisedBaselineCaseDetails,
  normalisedUpdated: normalisedUpdatedCaseDetails,
});

await testInfo.attach('civil case create flag data loss report.md', {
  body: report,
  contentType: 'text/markdown',
});
```

The important evidence is the changed-path summary and assertion result, not the full raw case body.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    await ensureSession(CIVIL_API_SETUP_ALIAS);
+    await ensureSession(COURT_STAFF_ALIAS);
+  });
+
+  test.beforeEach(async ({ page }, testInfo) => {
+    await retryOnTransientFailure(
+      async () => {
+        await ensureAuthenticatedPage(page, CIVIL_API_SETUP_ALIAS, { waitForSelector: 'exui-header' });
+
+        const setup = await createCivilLipCaseInMediationViaApi({
```
andyg-cgi, line 55: This passes `expectedState`, but the helper currently only logs that value and returns after the last business-process wait. That means the baseline can be captured before the case is actually in Mediation, so the test does not quite prove the contract in the PR description.

Can we hard-assert the state before taking the baseline, ideally by making the helper wait before it returns?

```ts
const setup = await createCivilLipCaseInMediationViaApi({
  expectedState: MEDIATION_STATE,
  page,
  useGeneratedUsers: true,
});

expect(resolveCcdCaseStateId(setup.caseDetails)).toBe(MEDIATION_STATE);
caseNumber = setup.caseNumber;
baselineCaseDetails = setup.caseDetails;
```

There is already a `waitForCaseState` pattern lower in this helper, so I would reuse that rather than relying on the later soft assertion after the flag event.

playwright_tests_new/E2E/page-objects/pages/exui/caseFlag.po.ts
```diff
@@ -0,0 +1,233 @@
...
+
+type CcdCaseTrigger = {
+  id?: string;
+  name?: string;
+  label?: string;
+  description?: string;
+};
+
+const CREATE_CASE_FLAG_ACTION_LABELS = ['Create case flag', 'Create a case flag', 'Create Case Flag', 'Create Case Flags'];
+const CREATE_CASE_FLAG_EVENT_ID_CANDIDATES = [
```
andyg-cgi: For this regression I think we need to be stricter about the event being exercised. The test/PR is about Civil `CREATE_CASE_FLAGS`, but this fallback list can silently move on to `manageFlags`, `createCaseFlag`, etc. and the success check only looks at the banner text.

Can we make the expected event explicit and fail if a different trigger is used?

```ts
const EXPECTED_CREATE_CASE_FLAG_EVENT_ID =
  process.env.PW_CIVIL_CREATE_CASE_FLAG_EVENT_ID?.trim() || 'CREATE_CASE_FLAGS';

// after resolving the trigger
if (triggerId !== EXPECTED_CREATE_CASE_FLAG_EVENT_ID) {
  throw new Error(
    `Expected Civil create-case-flag event '${EXPECTED_CREATE_CASE_FLAG_EVENT_ID}' but resolved '${triggerId}'`
  );
}
```

If the alternative IDs are genuinely valid for this journey, the test name/report should say which event contract is being covered.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      await caseDetailsPage.exuiSpinnerComponent.wait();
+    });
+
+    await test.step('Verify successful submission in UI', async () => {
+      await caseFlagPage.waitForCreateCaseFlagSuccess(caseNumber);
+    });
+
+    await test.step('Retrieve updated case data via API and compare with baseline', async () => {
+      const updatedCaseDetails = await fetchCaseDetailsViaApi(page, caseNumber);
+      const normalisedBaselineCaseDetails = normaliseCaseDataForDataLossComparison(baselineCaseDetails);
```
andyg-cgi, line 152: Before normalising away the created flag, can we first assert that the Claimant 1 party-level `Other` flag actually exists? At the moment a banner plus a normalised equality check can pass without proving the flag was persisted on the intended party.

Something close to the existing `caseFlags.positive.spec.ts` pattern would make this much stronger:

```ts
await caseDetailsPage.selectCaseDetailsTab('Flags');

await expect
  .poll(async () => {
    const partyFlagsTable = await caseDetailsPage.waitForTableByName(TEST_FLAG_COMMENT, {
      timeoutMs: 15_000,
    });
    const rows = await tableUtils.parseDataTable(partyFlagsTable);

    return rows.some((row) =>
      rowMatchesExpected(row, {
        'Party level flags': 'Other',
        Comments: TEST_FLAG_COMMENT,
        'Flag status': 'ACTIVE',
      })
    );
  }, { timeout: 60_000, intervals: [1_000, 2_000, 3_000] })
  .toBe(true);
```

Then the data-loss comparison is proving "the expected flag was added and nothing else changed", rather than only proving the normaliser can remove the flag entry.

yarn-audit-known-issues
```diff
@@ -27,6 +27,10 @@
...
 {"value":"inflight","children":{"ID":"inflight (deprecation)","Issue":"This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.","Severity":"moderate","Vulnerable Versions":"1.0.6","Tree Versions":["1.0.6"],"Dependents":["glob@npm:7.2.3"]}}
 {"value":"lodash.isequal","children":{"ID":"lodash.isequal (deprecation)","Issue":"This package is deprecated. Use require('node:util').isDeepStrictEqual instead.","Severity":"moderate","Vulnerable Versions":"4.5.0","Tree Versions":["4.5.0"],"Dependents":["@fast-csv/format@npm:4.3.5"]}}
+{"value":"mermaid","children":{"ID":1117991,"Issue":"Mermaid Gantt Charts are vulnerable to an Infinite Loop DoS","URL":"https://github.com/advisories/GHSA-6m6c-36f7-fhxh","Severity":"moderate","Vulnerable Versions":">=11.0.0-alpha.1 <=11.14.0","Tree Versions":["11.12.0"],"Dependents":["ngx-markdown@virtual:478250b179e2f7a41962cb81e8de022adafb1a3a18c5c9a01a14fbfc1b28d5290463c48c9e2b547a1f1c34dc9b7b468a7fcd7685a99bff9367385d59331a4cd4#npm:18.1.0"]}}
```
andyg-cgi: Can we keep these audit-baseline changes out of this E2E PR? The PR does not change `package.json` or `yarn.lock`, but it adds new Mermaid CVE suppressions and changes the Axios advisory entry.

I would either remove these lines from this PR, or split them into a dedicated audit PR with the usual context, for example:

```md
Advisory: GHSA-...
Package/version: mermaid 11.12.0
Reason for temporary suppression:
Mitigation/risk owner:
Validation run:
Follow-up ticket:
```

Changing the security waiver baseline without that provenance makes this much harder to approve.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+      `Failed to create Civil court staff account '${email}' through IDAM testing-support/accounts ` +
+        `(HTTP ${response.status()}). Roles='${roles.join(',')}'. Body='${body.slice(0, 500)}'. ` +
+        "Use a Civil admin role set that IDAM burner users accept, for example 'caseworker,caseworker-civil,caseworker-civil-admin,pui-case-manager'."
+    );
+  }
+
+  return { email, password };
+}
+
+export async function createCivilLipCaseInMediationViaApi(options: {
```
andyg-cgi, line 74: This helper has become a bit too broad for one file. It is creating users, resolving env config and internal URLs, handling IDAM/S2S, polling business processes, submitting Civil events and holding large payload builders. That makes the PR hard to review and leaves the most important support code without focused tests.

Can we split it into smaller modules before merging? A shape like this would be easier to review and cover:

```text
utils/test-setup/civil/
  civilApiConfig.ts
  civilRuntimeUsers.ts
  civilIdamTokens.ts
  civilEventApi.ts
  civilMediationJourney.ts
  civilPayloads.ts
```

The data-loss normaliser and journey setup should also have fake/unit coverage for the important cases, not only coverage via this one long E2E.

playwright_tests_new/E2E/page-objects/pages/exui/caseFlag.po.ts
```diff
@@ -0,0 +1,233 @@
...
+    await this.selectOtherFlagType();
+    await this.addCaseFlagCommentsIfRequested(comment);
+    await this.confirmActiveFlagStatusIfRequested();
+    await expect(
+      this.page.getByRole('heading', { name: /review flag details|check your answers|review details/i }).first()
+    ).toBeVisible();
+  }
+
+  async waitForCreateCaseFlagSuccess(caseNumber: string): Promise<void> {
+    await expect
```
andyg-cgi: This is a scenario verdict hidden inside the page object. I agree with the earlier comment that the page object should expose interactions/state, and the spec should own the `expect`.

For example:

```ts
// page object
async hasCreateCaseFlagSuccess(caseNumber: string): Promise<boolean> {
  await this.throwIfCreateCaseFlagErrorVisible();
  const bannerText = await this.caseAlertSuccessMessage.innerText().catch(() => '');
  return this.isCreateCaseFlagBannerVisible(bannerText, caseNumber);
}

// spec
await expect
  .poll(() => caseFlagPage.hasCreateCaseFlagSuccess(caseNumber), {
    timeout: 60_000,
    intervals: [1_000, 2_000, 3_000],
  })
  .toBe(true);
```

That keeps the business assertion visible in the spec and makes the page object easier to reuse.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  type CcdCaseDetails,
+} from '../../utils/test-setup/journeys/civilCaseJourneys';
+import { formatErrorMessage, isDependencyEnvironmentFailure, retryOnTransientFailure } from '../../utils/transient-failure.utils';
+
+const CIVIL_API_SETUP_ALIAS = getCivilCaseFlagsSetupAlias();
+const COURT_STAFF_ALIAS = getCivilCaseFlagsCourtStaffAlias();
+const MEDIATION_STATE = process.env.PW_CIVIL_MEDIATION_CASE_STATE?.trim() || 'IN_MEDIATION';
+const TEST_FLAG_COMMENT = 'Data loss Civil Create Case Flag';
+
+test.describe('Civil Create Case Flag data loss regression', { tag: ['@e2e', '@e2e-case-flags', '@e2e-civil-data-loss'] }, () => {
```
andyg-cgi: Given how much setup this scenario performs, I think we need an explicit decision on whether it belongs in the PR E2E path. Jenkins is already red and there is an existing comment asking whether this should only run nightly.

If the intent is nightly-only, I would tag it that way and make the routing obvious:

```ts
test.describe('Civil Create Case Flag data loss regression', {
  tag: ['@e2e', '@nightly', '@e2e-case-flags', '@e2e-civil-data-loss'],
}, () => {
```

If it is meant to stay in PR runs, can we add runtime evidence and deterministic user/setup guarantees so it does not become a slow/flaky merge gate?

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1684 @@
...
+const DEFAULT_STATE_WAIT_TIMEOUT_MS = 180_000;
+const DEFAULT_STATE_WAIT_INTERVAL_MS = 3_000;
+const DEFAULT_CASE_DETAILS_FETCH_WAIT_TIMEOUT_MS = 60_000;
+const DEFAULT_CIVIL_SERVICE_BUSINESS_PROCESS_WAIT_TIMEOUT_MS = 420_000;
+const DEFAULT_BUSINESS_PROCESS_WAIT_INTERVAL_MS = 3_000;
+const DEFAULT_CCD_EVENT_TRIGGER_WAIT_TIMEOUT_MS = 420_000;
+const DEFAULT_CCD_EVENT_TRIGGER_WAIT_INTERVAL_MS = 3_000;
+const DEFAULT_CIVIL_ROLE_ASSIGNMENT_WAIT_TIMEOUT_MS = 60_000;
+const DEFAULT_CIVIL_S2S_MICROSERVICE = 'civil_service';
+const DEFAULT_CIVIL_CLAIMANT_EMAIL = 'civilmoneyclaimsdemo@gmail.com';
```
andyg-cgi: Can we avoid defaulting to fixed live-looking Civil accounts here? The helper can generate users, but these constants still mean the suite can fall back to shared/static identities depending on which password env vars are present.

I would rather make generated or explicitly configured users the only valid paths:

```ts
const claimantEmail = firstNonEmpty(
  process.env.PW_CIVIL_CLAIMANT_EMAIL,
  process.env.CLAIMANT_CITIZEN_EMAIL
);

if (!claimantEmail && !createCitizenAccounts) {
  throw new Error('Civil data-loss setup requires generated users or explicit Civil claimant credentials');
}
```

That would make CI failures easier to reason about and avoid logging/reusing shared account identifiers.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+test.describe('Civil Create Case Flag data loss regression', { tag: DATA_LOSS_TEST_TAGS }, () => {
+  test.describe.configure({ timeout: DATA_LOSS_TEST_TIMEOUT_MS });
+
+  let caseNumber: string;
+  let baselineCaseDetails: CcdCaseDetails;
+
+  test.beforeAll(async ({ browser }) => {
+    await configureCivilCaseFlagsRuntimeUsers(browser);
+    await ensureSession(CIVIL_API_SETUP_ALIAS);
+    await ensureSession(COURT_STAFF_ALIAS);
```
andyg-cgi, line 47: This setup session is still fragile. `CIVIL_API_SETUP_ALIAS` defaults to `CIVIL_SOLICITOR`, but that alias is not in the normal runtime-user map or static user config unless extra Civil env vars happen to be present. If they are not present, this can fail in `ensureSession` before the actual scenario starts.

As this setup path is creating the Civil case via API with generated users, I’d remove the setup UI session altogether, or explicitly register a real setup `SessionIdentity` before using it. For example, the `beforeAll` can be narrowed to the UI user that is actually needed:

```suggestion
  test.beforeAll(async ({ browser }) => {
    await configureCivilCaseFlagsRuntimeUsers(browser);
    await ensureSession(COURT_STAFF_ALIAS);
  });
```

Then the `beforeEach` can call `createCivilLipCaseInMediationViaApi(...)` directly without logging the page in as `CIVIL_SOLICITOR` first.

playwright_tests_new/E2E/tag-filter.json
```diff
@@ -3,6 +3,7 @@
   "availableTags": [
     "@e2e",
     "@nightly",
+    "@e2e-civil-data-loss",
```
andyg-cgi: Adding `@e2e-civil-data-loss` to `availableTags` is only half the routing. The spec also has `@nightly`, but this file still has `"excludedTags": []`, so the default PR E2E path can still pick the long Civil data-loss test when Jenkins passes empty include/exclude params.

I’d default-exclude this tag and let nightly include it deliberately:

```json
{
  "excludedTags": ["@e2e-civil-data-loss"],
  "availableTags": [
    "@e2e",
    "@nightly",
    "@e2e-civil-data-loss"
  ]
}
```

That keeps the expensive data-loss journey out of PR smoke by default without losing the nightly route.

playwright_tests_new/E2E/utils/case-flags.utils.ts
```diff
@@ -16,3 +23,293 @@ export function rowMatchesExpected(row: Record<string, string>, expected: Record
...
+}
+
+function collectChangedPaths(before: unknown, after: unknown, path = '$', changes: string[] = []): string[] {
+  if (changes.length >= 50 || Object.is(before, after)) {
+    return changes;
+  }
+
+  if (Array.isArray(before) || Array.isArray(after)) {
+    if (!Array.isArray(before) || !Array.isArray(after)) {
+      changes.push(`${path}: ${formatReportValue(before)} => ${formatReportValue(after)}`);
```
andyg-cgi: The raw JSON attachments are now local-only, which is good, but this report is still always attached and it still writes before/after values for changed paths. On the failure this test is designed to catch, that can put live case-data values into the CI report artefacts.

I’d make the default report path-only, and only include values behind the same local debug gate as the raw JSON attachments. Something like:

```ts
const includeChangedValues = rawCivilDataLossAttachmentsEnabled();
const changedPaths = collectChangedPaths(options.normalisedBaseline, options.normalisedUpdated, {
  includeValues: includeChangedValues,
});

function formatChangedPath(path: string, before: unknown, after: unknown, includeValues: boolean): string {
  return includeValues ? `${path}: ${formatReportValue(before)} => ${formatReportValue(after)}` : path;
}
```

The CI-safe evidence should be the changed path list and the failed assertion, not the field values themselves.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1741 @@
...
+    caseNumber,
+    claimantEmail: config.claimantUser.email,
+    defendantEmail: config.defendantUser.email,
+    expectedState,
+  });
+
+  return {
+    caseNumber,
+    caseDetails: claimantResponseDetails,
+  };
```
andyg-cgi, line 168: `expectedState` is still only logged by this helper, and the helper returns `claimantResponseDetails` from an earlier event rather than the case details after the final Mediation wait. The spec works around this by doing another wait, but the exported helper contract is misleading and easy to reuse incorrectly.

Can we make the helper return the state it promises? For example:

```suggestion
  const caseDetails = await waitForCaseState({
    page: options.page,
    caseNumber,
    expectedState,
    context: 'after Civil LiP mediation setup',
  });

  return { caseNumber, caseDetails };
```

I’d also add a focused fake/unit test for this contract so a future caller cannot get a pre-mediation payload by accident.

yarn-audit-known-issues
```diff
@@ -1,13 +1,12 @@
+{"value":"@nevware21/ts-utils","children":{"ID":1119462,"Issue":"@nevware21/ts-utils: Prototype Pollution in objDeepCopy/objCopyProps via for...in without hasOwnProperty","URL":"https://github.com/advisories/GHSA-x7j8-49r8-mr43","Severity":"high","Vulnerable Versions":"<=0.13.0","Tree Versions":["0.12.5","0.13.0"],"Dependents":["@microsoft/applicationinsights-web@virtual:478250b179e2f7a41962cb81e8de022adafb1a3a18c5c9a01a14fbfc1b28d5290463c48c9e2b547a1f1c34dc9b7b468a7fcd7685a99bff9367385d59331a4cd4#npm:3.3.10","@nevware21/ts-async@npm:0.5.4"]}}
```
andyg-cgi: I would take the audit baseline changes out of this E2E PR. There are no matching `package.json` or `yarn.lock` changes here, but this file adds/re-keys several advisories and is also the file that currently conflicts with `master`, which is why Jenkins reports that this commit cannot be built.

For this PR I’d restore the audit file to `master` and handle the baseline refresh separately. If the refresh is needed, it should come with provenance in its own PR, for example:

```md
Audit baseline refresh

Command: yarn npm audit --recursive --environment production --json > yarn-audit-known-issues
Package/advisory changes: GHSA-...
Reason for temporary suppression:
Risk owner:
Follow-up ticket:
Validation evidence:
```

That keeps the data-loss E2E review separate from security-waiver changes and clears the current merge conflict.

andyg-cgi (changes requested):
Requesting changes for now. Re-checked current head `5d93bc5488ed8a49c5e146839898ec8db4eda595`; the PR is still not buildable because GitHub reports the branch as dirty and Jenkins `continuous-integration/jenkins/pr-head` is still erroring with “This commit cannot be built”. A local merge check also still conflicts in `yarn-audit-known-issues`.

The inline comments are the main things I’d fix before another review: remove/split the audit baseline change, keep the long Civil data-loss test out of default PR E2E, remove the fragile default setup UI session, make the markdown report CI-safe by default, and tighten the Civil mediation helper contract.

playwright_tests_new/E2E/utils/test-setup/journeys/civilCaseJourneys.ts
```diff
@@ -0,0 +1,1778 @@
...
+      `Failed to create Civil court staff account '${email}' through IDAM testing support ` +
+        `(HTTP ${response.status()}). Roles='${roles.join(',')}'. Body='${body.slice(0, 500)}'. ` +
+        "Use a Civil admin role set that IDAM burner users accept, for example 'caseworker,caseworker-civil,caseworker-civil-admin,pui-case-manager'."
+    );
+  }
+
+  return { email, password };
+}
+
+export async function createCivilLipCaseInMediationViaApi(options: {
```
andyg-cgi, line 74: This helper is doing a lot now: Civil API setup, generated users, IDAM account creation, token lookup, S2S, role assignment, business-process polling and the payload builders. I can follow why it grew this way, but it is becoming hard to review as one shared file.

I would either split it now or at least put fake-driven tests around the risky seams before we rely on it from more specs. A shape like this would make the contract easier to keep stable:

```text
utils/test-setup/civil/
  civilApiConfig.ts
  civilIdam.ts
  civilS2s.ts
  civilEventApi.ts
  civilMediationJourney.ts
  civilPayloads.ts
```

Even if the split is deferred, I would add focused unit coverage for the config resolution, generated-user path, role assignment retry, business-process polling and final-state return contract. The current unit coverage is good for the normaliser, but it does not really protect most of this orchestration.

playwright_tests_new/E2E/tag-filter.json
```diff
@@ -1,8 +1,9 @@
 {
-  "excludedTags": [],
+  "excludedTags": ["@nightly"],
```
andyg-cgi: I think this is now mostly doing the right thing by keeping the long data-loss journey out of the normal PR run, but I would make the evidence route really explicit before this is merged.

Because the spec has both `@nightly` and `@e2e-civil-data-loss`, and default E2E now excludes `@nightly`, the green PR build does not prove this new regression actually ran. I would add a targeted/nightly run note to the PR with the exact command/env used, for example:

```bash
E2E_PW_INCLUDE_TAGS=@e2e-civil-data-loss \
E2E_PW_EXCLUDED_TAGS_OVERRIDE=@none \
yarn test:playwrightE2E:raw
```

Or if the intention is strictly nightly only, link the nightly Jenkins/Odhín evidence from the current head. Otherwise this PR can look green while the main test it adds has not been exercised.

andyg-cgi:
I’ve had another pass over the current head. I’d still hold this one for changes, mainly because the shape is too heavy for the default PR E2E lane and one of the documented tag commands is currently broken.

The useful part of the PR is clear: it is trying to prove that creating a Civil case flag does not wipe existing case data. I think that is worth having. The problem is that the implementation adds a full Civil LiP setup stack, dynamic user creation, multiple Civil API transitions, role assignment, UI search, UI flag creation, flags-tab validation, and API comparison into a default `@e2e` PR test with a 35 minute timeout. That is a lot of downstream dependency pressure for every PR run.

### 1. Keep this out of default PR E2E

The spec is currently tagged like this:

```ts
const DATA_LOSS_TEST_TAGS = ['@e2e', '@e2e-case-flags', '@e2e-civil-data-loss'];
const DATA_LOSS_TEST_TIMEOUT_MS = resolvePositiveInt(process.env.PW_CIVIL_DATA_LOSS_TEST_TIMEOUT_MS, 35 * 60_000);
```

Given the setup cost and number of external dependencies, I’d make this a nightly/explicit run by default. Something along these lines would be easier to sign off:

```suggestion
const DATA_LOSS_TEST_TAGS = ['@e2e', '@nightly', '@e2e-case-flags', '@e2e-civil-data-loss'];
const DATA_LOSS_TEST_TIMEOUT_MS = resolvePositiveInt(process.env.PW_CIVIL_DATA_LOSS_TEST_TIMEOUT_MS, 35 * 60_000);
```

And update the README wording so it does not say this runs in the default PR/CNP E2E set:

```md
- The Civil data-loss regression is tagged `@e2e-civil-data-loss` and `@nightly`, so it is excluded from the default PR E2E set. Run it explicitly when validating the Civil Create Case Flag data-loss path:

```bash
E2E_PW_INCLUDE_TAGS=@e2e-civil-data-loss \
E2E_PW_EXCLUDED_TAGS_OVERRIDE=@none \
yarn test:playwrightE2E:raw
```
```

If the intent really is to run this on every PR, then I’d expect much stronger evidence that it is stable at PR frequency and does not materially slow the lane. Right now the design looks like nightly/regression coverage, not default PR coverage.

### 2. Register the new tag properly

The README documents:

```bash
E2E_PW_INCLUDE_TAGS=@e2e-civil-data-loss yarn test:playwrightE2E:raw
```

but `playwright_tests_new/E2E/tag-filter.json` does not include `@e2e-civil-data-loss` in `availableTags`. The config validates include tags against that catalogue, so the documented command will fail before it gets to the test.

Suggested fix:

```suggestion
    "@e2e-case-flags",
    "@e2e-civil-data-loss",
    "@e2e-create-case",
```

I’d also add it to the config coverage assertion in `playwright_tests_new/api/playwright-config-coverage.api.ts`:

```ts
expect(filters.availableTags).toEqual(
  expect.arrayContaining([
    '@e2e-case-file-view',
    '@e2e-case-flags',
    '@e2e-civil-data-loss',
    '@e2e-create-case',
    '@e2e-data-loss',
    '@e2e-document-upload',
    '@e2e-document-upload-v1',
    '@e2e-manage-tasks',
    '@e2e-manage-tasks-assigned',
    '@e2e-media-viewer',
    '@e2e-search-case',
    '@e2e-update-case',
  ])
);
```

### 3. Tighten the page object instead of broad polling/fallbacks

`CaseFlagPage` is doing quite a lot of defensive probing and “first visible matching radio” behaviour. This is the part I’d expect to become flaky or to hide a real page-shape change.

Current pattern:

```ts
private async selectFirstVisibleRadio(labels: RegExp[], context: string): Promise<void> {
  const deadline = Date.now() + 60_000;

  while (Date.now() < deadline) {
    for (const label of labels) {
      const radio = this.page.getByLabel(label).first();
      if (await radio.isVisible().catch(() => false)) {
        await radio.check();
        return;
      }
    }

    await this.page.waitForTimeout(500);
  }

  // label scrape...
}
```

I’d rather make the Civil wizard contract explicit and fail loudly if the expected shape is not there:

```ts
private async checkRadio(label: RegExp, context: string): Promise<void> {
  const radio = this.page.getByLabel(label).first();
  await radio.waitFor({ state: 'visible', timeout: 30_000 });
  await radio.check();
}

private async selectPartyFlagLocationForClaimant1(): Promise<void> {
  await this.checkRadio(/^Party$/i, 'party flag location level');
  await this.clickCaseFlagNext();

  await this.checkRadio(/^Claimant 1(?:\s*\(.*\))?$/i, 'Claimant 1 party flag location');
  await this.clickCaseFlagNext();
}

private async selectOtherFlagType(): Promise<void> {
  await this.checkRadio(/^Other$/i, 'Other flag type');
  await this.clickCaseFlagNext();

  const customFlagTypeInput = this.page.getByLabel(/Enter a flag type|Other description/i).first();
  await customFlagTypeInput.waitFor({ state: 'visible', timeout: 30_000 });
  await customFlagTypeInput.fill('Other');
  await this.clickCaseFlagNext();
}
```

If any of those pages is genuinely optional in Civil, that should be captured as a named branch with evidence, not by silently skipping controls when they are absent.

### 4. Add direct tests for the API setup seams

The new Civil setup support is now doing important orchestration: S2S token creation, IDAM account creation/fallbacks, Civil event submission, CCD validation, role assignment, and business process polling. The added unit tests cover some useful pure/helper behaviour, but not enough of the failure-prone seams.

I’d add fake-driven tests around at least these:

```ts
test('submitCivilServiceEvent sends IDAM and S2S headers and returns the case payload', async () => {
  const requests: Array<{ url: string; headers: Record<string, string>; data: unknown }> = [];
  const page = {
    request: {
      post: async (url: string, options: any) => {
        requests.push({ url, headers: options.headers, data: options.data });
        return {
          ok: () => true,
          status: () => 200,
          json: async () => ({ id: '1234567890123456', state: 'IN_MEDIATION' }),
        };
      },
    },
  } as never;

  const result = await civilCaseJourneyTestHelpers.submitCivilServiceEventForTest({
    page,
    url: 'https://civil-service.example.test/cases/123/citizen/user/event',
    eventPayload: { event: 'CLAIMANT_RESPONSE_CUI' },
    idamToken: 'idam-token',
    s2sToken: 's2s-token',
    context: 'submit claimant response',
  });

  expect(result).toEqual({ id: '1234567890123456', state: 'IN_MEDIATION' });
  expect(requests[0].headers).toMatchObject({
    Authorization: 'Bearer idam-token',
    ServiceAuthorization: 's2s-token',
    'Content-Type': 'application/json',
  });
});
```

```ts
test('waitForFinishedCivilBusinessProcess fails fast on incident message', async () => {
  const page = {
    waitForTimeout: async () => undefined,
    request: {
      get: async () => ({
        ok: () => true,
        status: () => 200,
        json: async () => ({
          businessProcess: { status: 'STARTED' },
          incidentMessage: 'Camunda failed',
        }),
      }),
    },
  } as never;

  await expect(
    civilCaseJourneyTestHelpers.waitForFinishedCivilBusinessProcess(
      page,
      { civilServiceUrl: 'https://civil-service.example.test' } as never,
      'idam-token',
      '1234567890123456',
      { timeoutMs: 1_000 }
    )
  ).rejects.toThrow('Civil business process failed for case 1234567890123456: Camunda failed');
});
```

The helper exports may need a small `__test__` surface for this, but I’d keep that preferable to leaving this much orchestration only indirectly covered by the browser scenario.

### 5. Redact configured account identifiers in errors

A few error messages include full configured emails and sliced dependency response bodies. That is useful locally, but risky in CI logs/reports when the helper can use configured accounts.

For example, instead of:

```ts
throw new Error(
  `Failed to get IDAM token for '${user.email}' (HTTP ${response.status()}). Body='${body.slice(0, 500)}'`
);
```

I’d use a small redaction helper and keep raw bodies behind an explicit debug flag:

```ts
function redactEmail(value: string): string {
  return value.replace(/^(.).+(@.+)$/, '$1***$2');
}

function dependencyBodyForError(body: string): string {
  return process.env.PW_CIVIL_DEBUG_DEPENDENCY_BODIES === 'true'
    ? ` Body='${body.slice(0, 500)}'`
    : '';
}
```

Then:

```ts
const body = await response.text().catch(() => '');
throw new Error(
  `Failed to get IDAM token for '${redactEmail(user.email)}' (HTTP ${response.status()}).${dependencyBodyForError(body)}`
);
```

That keeps normal CI diagnostics useful without putting real account identifiers or raw service payloads into reports by default.

### Bottom line

I think the test objective is good, but I would not merge this in its current shape. Minimum for me would be:

- keep the Civil data-loss test out of default PR E2E, or provide strong evidence that it is PR-lane safe
- register `@e2e-civil-data-loss` in `tag-filter.json` and config coverage
- tighten `CaseFlagPage` so it proves the expected wizard instead of probing/skipping broadly
- add direct unit coverage for the Civil API orchestration failure paths
- redact configured identifiers and raw dependency bodies in CI-facing errors

playwright_tests_new/api/unit/case-flags-data-loss.unit.api.ts
```diff
@@ -0,0 +1,271 @@
...
+      },
+    ],
+  };
+}
+
+function configureRequiredCivilMediationEnv(): void {
+  for (const [key, value] of Object.entries(REQUIRED_CIVIL_MEDIATION_ENV)) {
+    process.env[key] = value;
+  }
+}
```
danH-test: If possible keep test functions out of tests, to reduce duplication and keep tests readable.

playwright_tests_new/E2E/page-objects/pages/exui/caseFlag.po.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  private async clickCaseFlagNext(): Promise<void> {
+    const nextButton = this.page.getByRole('button', { name: /^(Next|Continue)$/i }).first();
+    await nextButton.waitFor({ state: 'visible', timeout: 30_000 });
+    await nextButton.click();
+    await this.page
+      .locator('xuilib-loading-spinner')
+      .first()
+      .waitFor({ state: 'hidden', timeout: 30_000 })
+      .catch(() => undefined);
```
danH-test: There is already a spinner wait function in the common lib, consider referencing that here

danH-test (changes requested):

playwright_tests_new/E2E/page-objects/pages/exui/caseFlag.po.ts
```diff
@@ -0,0 +1,227 @@
...
+  private async hasCallbackValidationErrorAlert(timeoutMs = 1_000): Promise<boolean> {
+    return this.page
+      .getByRole('alert')
+      .filter({ hasText: /callback data failed validation/i })
+      .first()
+      .isVisible({ timeout: timeoutMs })
+      .catch(() => false);
+  }
+
+  private async selectFirstVisibleRadio(labels: RegExp[], context: string): Promise<void> {
```
andyg-cgi: This is much closer now, and I like that the spec now proves the final Flags tab row and then compares the case data via API. The remaining thing I would tighten is the wizard contract itself. At the moment this helper still does a broad "try whichever matching radio becomes visible" loop, and a few later steps silently continue when the expected Civil page/control is not visible quickly. That makes the test more forgiving than the user journey it is trying to prove.

I do not think this needs a big rewrite. I would make the known Civil path explicit, then only add named optional branches if we have evidence that Civil genuinely renders more than one valid shape. Something like this would make failures easier to diagnose and would avoid hiding a page-shape change behind a successful final assertion:

```ts
private async checkRadio(label: RegExp, context: string): Promise<void> {
  const radio = this.page.getByLabel(label).first();
  await radio.waitFor({ state: 'visible', timeout: 30_000 });
  await radio.check();
}

private async fillVisibleTextArea(label: RegExp, value: string, context: string): Promise<void> {
  const input = this.page.getByLabel(label).first();
  await input.waitFor({ state: 'visible', timeout: 30_000 });
  await input.fill(value);
}

private async selectPartyFlagLocationForClaimant1(): Promise<void> {
  await this.checkRadio(/^Party$/i, 'case flag location level');
  await this.clickCaseFlagNext();

  await this.checkRadio(/^Claimant 1(?:\s*\(.*\))?$/i, 'Claimant 1 party flag location');
  await this.clickCaseFlagNext();
}

private async selectOtherFlagType(): Promise<void> {
  await this.checkRadio(/^Other$/i, 'Other flag type');
  await this.clickCaseFlagNext();

  await this.fillVisibleTextArea(/Enter a flag type|Other description/i, 'Other', 'custom Other flag description');
  await this.clickCaseFlagNext();
}

private async addCaseFlagComments(comment: string): Promise<void> {
  const comments = this.page.locator('ccd-add-comments textarea, textarea#flagComments, textarea').first();
  await comments.waitFor({ state: 'visible', timeout: 30_000 });
  await comments.fill(comment);
  await this.clickCaseFlagNext();
}

private async confirmActiveFlagStatus(): Promise<void> {
  await this.checkRadio(/^Active$/i, 'case flag active status');

  const reason = this.page.getByLabel(/Describe reason/i).first();
  if (await reason.isVisible({ timeout: 1_000 }).catch(() => false)) {
    await reason.fill('Created for data loss regression coverage');
  }

  await this.clickCaseFlagNext();
}
```

If any of these pages really is optional for this Civil event, I would keep that as an explicit method/branch with a short comment explaining the product shape rather than letting the helper silently skip it. That keeps the test friendly to future maintainers without making the scenario heavier.

playwright_tests_new/E2E/utils/test-setup/journeys/civilEventApi.ts
```diff
@@ -0,0 +1,344 @@
...
+    data: eventBody,
+    failOnStatusCode: false,
+    headers: CCD_CREATE_EVENT_HEADERS,
+    timeout: resolveCivilApiRequestTimeoutMs(),
+  });
+  if (!submitResponse.ok()) {
+    const body = await submitResponse.text().catch(() => '');
+    throw new Error(
+      `Civil case setup failed to submit event '${options.eventId}' (HTTP ${submitResponse.status()}). ` +
+        `Path='${submitPath}'. Body='${body.slice(0, 500)}'`
```
andyg-cgi: Nice improvement on the IDAM/S2S paths where the error body is now behind `PW_CIVIL_DEBUG_DEPENDENCY_BODIES`. I think it is worth applying the same pattern across the remaining new Civil helper errors so the normal CI/nightly failure output stays safe by default. This is not a big architectural ask; it is just making the diagnostics consistent.

There are still a few places in the new Civil setup surface that include `body.slice(0, 500)` directly in thrown errors, for example this submit path, the event-trigger fetch/validate paths, role assignment, case-details fetch, and the legacy IDAM fallback. Those upstream bodies are useful locally, but in CI reports they can contain dependency payloads that are noisy at best and sensitive at worst.

A small shared helper would keep the useful debug option without repeating the same guard in every file:

```ts
// civilDiagnostics.ts
export function dependencyBodyForError(body: string): string {
  if (process.env.PW_CIVIL_DEBUG_DEPENDENCY_BODIES === 'true') {
    return ` Body='${body.slice(0, 500)}'.`;
  }

  return '';
}

export function civilDependencyError(message: string, body: string): string {
  return `${message}${dependencyBodyForError(body)}`;
}
```

Then the failing submit path can stay just as diagnostic locally, while being quiet by default in CI:

```ts
if (!submitResponse.ok()) {
  const body = await submitResponse.text().catch(() => '');
  throw new Error(
    civilDependencyError(
      `Civil case setup failed to submit event '${options.eventId}' ` +
        `(HTTP ${submitResponse.status()}). Path='${submitPath}'.`,
      body
    )
  );
}
```

And the same helper could be used for the other new paths that currently do `Body='${body.slice(0, 500)}'`. That keeps the PR focused and avoids introducing a larger redaction framework just for test support.

playwright_tests_new/api/unit/case-flags-data-loss.unit.api.ts
danH-test: Re-reviewing, can still see these test functions in with the test.

playwright_tests_new/E2E/test/dataLossCivilCasesCreateCaseFlags/dataLossCivilCasesCreateCaseFlags.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+
+      expect(normalisedUpdatedCaseDetails).toEqual(normalisedBaselineCaseDetails);
+    });
+  });
+});
+
+function resolvePositiveInt(value: string | undefined, fallback: number): number {
+  const parsed = Number.parseInt(value ?? '', 10);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
```
danH-test: we still have a function in the spec file

andyg-cgi (dismissed):

danH-test (approved):

RiteshHMCTS (approved):

PR #5148, opened 2026-05-15
pofcc-131

anthonydummer:
### Jira link

https://tools.hmcts.net/jira/browse/POFCC-131

### Change description

Created a new function to be called when checking config roles for compliance.  If an AM roles matches a user role, further checks are made to ensure the role is a supported role.

### Dependency PRs

none

### Testing done

This has been tested locall.  New logger calls have been added to give clarity to testers which role is being used for a particular item.

- [ ] Yes
- [x] No

### Checklist


src/work-allocation/services/wa-supported-role-details.service.ts
```diff
@@ -0,0 +1,17 @@
+import { HttpClient } from '@angular/common/http';
```
Josh-HMCTS: I'm thinking that this may be better stored as part of the src/app/services/ccd-config/ccd-case.config.ts. this allows us to pass config around the app and as the data from these req remains static per deployment, doesn't seem right that we keep calling a backend api when we can effectively store it in the front end in the above config service. 
Unfortunately i think the case config links to toolkit so there may be changes needed in there.

connorpgpmcelroy (dismissed):
Difficult change to understand and I had to examine the code and the ticket for a while trying to figure out why this change was needed but think I agree with the changes made based on the ask (although not necessarily sure why the preview yaml has changed but there could be plenty of reasons for that). As someone who has more knowledge of WA and is often asked about it, it is in my interest to understand this fully. Therefore I would appreciate a demo or discussion about this at some point with me

api/configuration/menuConfigs/aat-diffs.ts
```diff
@@ -1,28 +1,10 @@
...
-    {
-      roles: ['caseworker-sscs-judge', 'caseworker-sscs-panelmember'],
-      text: 'My work',
-    },
-    {
-      roles: ['caseworker-sscs-judge', 'caseworker-sscs-panelmember'],
-      text: 'Search',
-    },
-  ],
+  '(judge)|(judiciary)|(panelmember)': [],
```
Josh-HMCTS: If this isnt needed i think it can probably just be removed

api/configuration/menuConfigs/aat-diffs.ts
```diff
@@ -1,28 +1,10 @@
...
-    {
-      roles: ['caseworker-sscs-clerk', 'caseworker-sscs-registrar'],
-      text: 'My work',
-    },
-    {
-      roles: ['caseworker-sscs-clerk', 'caseworker-sscs-registrar', 'caseworker-probate'],
-      text: 'Search',
-    },
-  ],
+  '.+': [],
```
Josh-HMCTS: same with this

Josh-HMCTS (dismissed):

api/configuration/menuConfigs/aat-diffs.ts
anthonydummer: Have removed caseworker-befta_master from base-config and added in here, so no longer empty

api/configuration/menuConfigs/aat-diffs.ts
anthonydummer: I have removed this

RiteshHMCTS (dismissed):

Josh-HMCTS (dismissed):

connorpgpmcelroy (dismissed):

PR #5152, opened 2026-05-15
#4505 | Text spacing Issues on Terms and Conditions Tab

renatasd-cgi:
### Jira link

See [EXUI-4505](https://tools.hmcts.net/jira/browse/EXUI-4505)

### Change description

This PR fixes unwanted whitespace added by Prettier in the Angular HTML template.

The long inline translation string inside the interpolation was replaced with a variable {{ key }} to avoid formatting issues and improve readability.


src/app/containers/legacy-terms-and-conditions/legacy-terms-and-conditions.component.ts
```diff
@@ -5,4 +5,55 @@ import { Component } from '@angular/core';
   selector: 'exui-legacy-terms-and-conditions',
   templateUrl: './legacy-terms-and-conditions.component.html',
 })
-export class LegacyTermsAndConditionsComponent {}
+export class LegacyTermsAndConditionsComponent {
+  public readonly introWithPrivacyPolicyText =
+    'By accessing this service (‘the Service’), you’re agreeing to the following terms of use. This includes the %privacyPolicyLink%.';
+  public readonly privacyPolicyLinkHtml = '<a href="/privacy-policy">privacy policy</a>';
+  public readonly accessingServiceIntroKey =
```
connorpgpmcelroy, line 12: Unsure why there is a difference between key and text here (with other component usage). Could they not all be text - key doesn't seem right in this instance

connorpgpmcelroy (commented):
Suggested change

src/app/containers/legacy-terms-and-conditions/legacy-terms-and-conditions.component.ts
renatasd-cgi: I used key here because the full text version causes Prettier to reformat the template and introduce extra whitespace in the rendered HTML. Shortening it was mainly to avoid that formatting issue and keep the output stable.

I agree text would be more consistent semantically, but in this case it changes the rendered output because of the formatter behavior. also it only cause it when using {{}} that's why it work on other pages.

connorpgpmcelroy (dismissed):
LGTM

connorpgpmcelroy (dismissed):

chrisjones-hmcts (dismissed):

RiteshHMCTS (approved):

chrisjones-hmcts (approved):

PR #5159, opened 2026-05-18
pofcc-75 - ensure focus is always set to the top of the page when navigating through the case management

jamescatto-cgi:
### Jira link

https://tools.hmcts.net/jira/browse/POFCC-75

### Change description

Ensure that there is always a main element with the ID of `content` to ensure the `skip to main content` link will have a valid target in the DOM.

Also, use the new `FocusService` in the `ccd-case-ui-toolkit` to tag an empty div element above the `skip to main content` link. This allows components within `ccd-case-ui-toolkit` to trigger a focus event on this element, which in turn allows the user to use the `tab` key to navigate to the `skip to main content` link. 

> N.B. It was not possible to focus on the skip link itself since that link is hidden and styled to become visible when it is given focus. Therefore, each page would always have this link in view.

There are no functional changes except that the tab ordering changed so that the initial focus for new pages is always above the `skip to main content` link.

### Dependency PRs

The corresponding change in the case ui library: https://github.com/hmcts/ccd-case-ui-toolkit/pull/2166

### Testing done

Added a unit test for the case home component listening on the router event.

- [ ] Yes
- [X] No


connorpgpmcelroy (dismissed):

anthonydummer (dismissed):

src/app/components/header/header.component.spec.ts
```diff
@@ -8,6 +8,7 @@ import { RpxTranslationService } from 'rpx-xui-translation';
 import { of } from 'rxjs';
 import { HeaderComponent } from './header.component';
 import { provideHttpClient, withInterceptorsFromDi } from '@angular/common/http';
+import { FocusService } from '@hmcts/ccd-case-ui-toolkit';
```
connorpgpmcelroy, line 11: Probably didn't pick up on this before but external imports should ideally be above internal imports

connorpgpmcelroy (dismissed):
Approved with comment

anthonydummer (dismissed):

PR #5160, opened 2026-05-18
EXUI-4585

Josh-HMCTS:
### Jira link

See [EXUI-4585](https://tools.hmcts.net/jira/browse/EXUI-4585)

### Change description

Change the current implementation of the caseWorkerCache to use a reduced set of role-names for searching on to reduce RAS load

Implement functionality to support users being searchable on all their available role_categories across all their jurisdictions

Fix an issue where the cache refresh call could be made multiple times if multiple users request the cache when the TTL is expired 

### Testing done

Ran code locally and tested against users with multiple role_categories and services, they remain searchable on WA related search fields


olusegz07 (approved):

RiteshHMCTS (dismissed):

RiteshHMCTS (approved):

balajisridharanhmcts (approved):
LGTM

PR #5246, opened 2026-06-23
Task/ex UI 4758 multiple role categories

connorpgpmcelroy:
### Jira link

https://tools.hmcts.net/jira/browse/EXUI-4758

### Change description

Allow use of multiple role categories within caseworkers and within the user (to allow searching for users with multiple role categories)

### Dependency PRs

https://github.com/hmcts/rpx-xui-common-lib/pull/548
https://github.com/hmcts/ccd-case-ui-toolkit/pull/2198

### Testing done

Confirmed this works for find person across pages (matches master functionality with addition of searching across multiple role categories)

- [ ] Yes
- [ x ] No


RiteshHMCTS (dismissed):

src/role-access/containers/allocate-role/allocate-role-home/allocate-role-home.component.ts
```diff
@@ -198,36 +198,28 @@ export class AllocateRoleHomeComponent implements OnInit, OnDestroy {
...
-                switch (this.userRole) {
-                  case UserRole.LegalOps:
-                    this.store.dispatch(new fromFeature.AllocateRoleChangeNavigation(AllocateRoleState.CHOOSE_ROLE));
-                    break;
-                  case UserRole.Judicial:
-                    this.store.dispatch(new fromFeature.AllocateRoleChangeNavigation(AllocateRoleState.CHOOSE_ALLOCATE_TO));
-                    break;
-                  default:
-                    throw new Error('Invalid user role');
+                if (this.hasUserRole(UserRole.Judicial)) {
```
Josh-HMCTS, line 201: I wonder if in this situation it could cause issues, for example if a user has selected under the legal ops option, but the user has the judicial role they would in effect be treated as judicial as it would be the first switch case hit?

src/work-allocation/components/case-manager-filter/case-manager-filter.component.ts
```diff
@@ -156,8 +156,9 @@ export class CaseManagerFilterComponent implements OnInit, OnDestroy {
 
   public ngOnInit(): void {
     this.appStoreSub = this.appStore.pipe(select(fromAppStore.getUserDetails)).subscribe((userDetails) => {
-      const isLegalOpsOrJudicialRole = AppUtils.getUserRole(userDetails?.userInfo?.roles || []);
-      const roleType = AppUtils.convertDomainToLabel(isLegalOpsOrJudicialRole);
+      const userRoleNames = AppUtils.getUserRoleNames(userDetails?.userInfo?.roles || []);
+      // EXUI-4758 - Get first user role category, otherwise default to ALL
+      const roleType = AppUtils.convertDomainToLabel(userRoleNames[0] || PersonRole.ALL);
```
Josh-HMCTS, line 161: I have some concern that defaulting to the first role could cause issue somewhere in the journeys. for example if a user chooses a ctsc role and the user has that in the role array, but if they take the first which is admin, not sure if that could cause issue.

src/work-allocation/containers/all-work-task/all-work-task.component.ts
```diff
@@ -100,7 +100,9 @@ export class AllWorkTaskComponent extends TaskListWrapperComponent {
       if (!userInfo) {
         return;
       }
-      const userRole: UserRole = AppUtils.getUserRole(userInfo.roles);
+      const userRoleNames: UserRole[] = AppUtils.getUserRoleNames(userInfo.roles);
+      // EXUI-4758 - Set to first user role category, otherwise default to LegalOps
+      const userRole: UserRole = userRoleNames[0] || UserRole.LegalOps;
```
Josh-HMCTS, line 105: Same as above with defaulting to [0] role

Josh-HMCTS (changes requested):
Might be worth catching up on the use of the defaulting to [0] role and what impacts this could potentially have

src/role-access/containers/allocate-role/allocate-role-home/allocate-role-home.component.ts
connorpgpmcelroy: So what happens is roleCategory check first i.e. the roleCategory of the role being allocated, then the userRole check i.e. does the user have this role? If the user has the judicial role but the user has selected legal ops the first check is legal ops and vice versa. Therefore choosing who to allocate will only happen if the user has the role of the roleCategory required and there are no problems here

src/work-allocation/components/case-manager-filter/case-manager-filter.component.ts
connorpgpmcelroy: It shouldn't cause an issue, at least not a functional one. This is just setting the default roleCategory for a user for the filter so the user must have a roleCategory regardless that will be correctly set. The only problem is if they are expecting a service specific roleCategory e.g. they are a CTSC on Civil and an admin on IA, then it may present a deviation from what they might be expecting. However there is no way to fix this unless, as I suggested we could do, we have service related role categories in the list. My thoughts are that this will add further complications and checks that we don't need when we're just talking about a default roleCategory on a filter that can be changed anyway.

src/work-allocation/containers/all-work-task/all-work-task.component.ts
connorpgpmcelroy: Again, it's only relevant to the filter default roleCategory so point above still stands.

Josh-HMCTS (dismissed):

PR #5268, opened 2026-06-30
initial commit to remove specific access roles for config checks

anthonydummer:
### Jira link

https://tools.hmcts.net/jira/browse/COT-1271

### Change description

Where the user roles are to be evaluated for the presence of the text judge or judicial in order to define the theme, or set the menu items a new function will remove the specific-access-approver- roles prior to the evaluation.  This is because there has always been a relationship between the approver and the role, so judicial users will have specific-access-approver-judicial approver roles.  However, this relationship is no longer being enforced and CTSC roles will now be granted specific-access-approver-judicial roles due to the fact that they do not have judicial roles.  The removal of the roles is therefore necessary as they do not drive the access they are being checked for but there presence is creating false judicial themes. 

### Testing done

This has been tested on the ctsc with this judicial approver role.  The user now correctly does not get the judicial theme. 

- [ ] Yes
- [x] No

### Checklist


src/app/containers/app-header/app-header.component.ts
```diff
@@ -126,10 +127,9 @@ export class AppHeaderComponent implements OnInit, OnDestroy {
 
   public setApplicationThemeForUser(): void {
     const availableThemes = environment.themes;
-    const userRoles = this.userRoles;
-    const userTheme = Object.keys(availableThemes).find((themeRegex) =>
-      userRoles.some((role) => new RegExp(themeRegex).test(role))
-    );
+    const userRoles = getUserRolesExcludingSpecificAccessApprover(this.userRoles);
+    const userTheme =
+      Object.keys(availableThemes).find((themeRegex) => userRoles.some((role) => new RegExp(themeRegex).test(role))) || '.+';
```
AntonyLeons, line 132: do we need this change?

AntonyLeons (approved):

src/app/containers/app-header/app-header.component.ts
kiran-yenigala-hmcts, line 132: can we keep the previous code?

const userTheme = Object.keys(availableThemes).find((themeRegex) =>
      userRoles.some((role) => new RegExp(themeRegex).test(role))
    );

kiran-yenigala-hmcts (changes requested):

src/app/containers/app-header/app-header.component.ts
anthonydummer: The theory of the fallback is if the user only has specific access roles (unlikely, possibly impossible), then getUserRolesExcludingSpecificAccessApprover() would return an empty array which would result in no theme being set.  So, in that scenario, it is just setting the default theme.
Happy to take this out if you would rather.

kiran-yenigala-hmcts (approved):

connorpgpmcelroy (approved):

PR #5306, opened 2026-07-15
Exui 4970

RiteshHMCTS:
### Jira link
https://tools.hmcts.net/jira/browse/EXUI-4970
https://tools.hmcts.net/jira/browse/EXUI-3289


Josh-HMCTS (approved):

connorpgpmcelroy (changes requested):
Don't think we should be going back to 2.4.4 here, we need to be on 2.4.5 right?

connorpgpmcelroy (approved):
downgrade explained to me

PR #5305, opened 2026-07-15
POFCC-365 decentralised ExUI

jamescatto-cgi:
### Jira link

https://tools.hmcts.net/jira/browse/POFCC-365

### Change description

The original [PR](https://github.com/hmcts/rpx-xui-webapp/pull/4943) got merged into an SRT feature branch. This PR brings it back

### Security Vulnerability Assessment

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?

- [ ] Yes
- [ ] No

### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change
- [ ] Can this PR be released on a Friday (ie: no functional code changes)

PR #5309, opened 2026-07-16
POFCC-329 - allow navigation links in the header to go to decentralised service URLs

jamescatto-cgi:
### Jira link

https://tools.hmcts.net/jira/browse/POFCC-329

### Change description

Using part of the decentralise mechanism of [pofcc-365](https://github.com/hmcts/rpx-xui-webapp/pull/5305) allow links in the header to open external service URLs.

The required config is:

* adding a new DECENTRALISED_SERVICE_MAP variable in `config/default.json`
* adding a new menu item in ExUI at `api/configuration/menuConfigs/base-config.ts`
    * the menu items are grouped into roles - only users with that role will see the menu items
    * the menu item should probably be added to the `pui-case-manager` role section - PCS users need to have the `pui-case-manager` role

For example:

In `config/default.json`:

```json
"decentralisedServiceMap": {    
  "PCS": { "id": "PCS", "baseUrl": "https://www.google.com" }  
}, 
```

In `api/configuration/menuConfigs/base-config.ts`:

```json
"(pui-case-manager)": [
  ...
  {      
    active: false,      
    href: '/duty-advisor',      
    roles: ['caseworker-pcs-solicitor'],      
    text: 'Duty Advisor',      
    decentralisedServiceId: 'PCS',    
  }
]
```

> N.B. the `decentralisedServiceId` must match the key in the `decentralisedServiceMap`.

> This configuration is not included in this PR. When the service is ready, the configuration can be added separately

Additionally, this PR includes several commits at the end that move the decentralisation code into a single package `src/decentralisation`. This will help encapsulate the code and simplify development. Only minor changes to type names were made and no functional changes have been made. There is still some duplication across the `api` and `src` modules where the `noc` changes require some of the decentralised code. This should be addressed in a future PR.

- [ ] Yes
- [ ] No

### Checklist


src/app/models/theming.model.ts
```diff
@@ -18,4 +18,5 @@ export interface NavigationItem {
   notFlags?: FlagDefinition[];
   ngClass?: string;
   align?: 'right' | 'left';
+  decentralisedServiceId?: string;
```
jamescatto-cgi, line 21: should the service ID be mapped here? Is it possible to get the service ID from the user session instead to avoid hardcoding the service ID here?

src/decentralisation/decentralised-casetype.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * These allow services to configure the URL paths into their application for specific case types.
+ * N.B. These definitions are duplicated in api/noc/index.ts and must correlate
+ * TODO created shared module that holds these definitions and any shared code
```
jamescatto-cgi, line 4: I have not done that in this PR to avoid complicating this PR. This can be done in a separate PR for clarity.

src/app/models/theming.model.ts
Josh-HMCTS: I'm not sure if its possible to easily retrieve the serviceID for the user, from a quick look i cant see it included in any of the data we already store that relates to a user. It may be possible but not sure off top of my head which endpoint/place we could get from and append to existing ngrx/storage.
Also a user could be part of multiple services so may be hard to determine which goes with which, without additional logic to try figure this out. 
I think in this case it may be better hardcoded as part of the config for the header?

src/decentralisation/decentralised-casetype.ts
Josh-HMCTS: Definitely agree a shared module would be beneficial, especially if as the todo mentions its duplicated on NOC already.

src/app/services/user/user.service.ts
```diff
@@ -1,15 +1,27 @@
...
+  constructor(
+    private readonly http: HttpClient,
+    private readonly sessionStorageService: SessionStorageService
+  ) {}
 
   public getUserDetails(refreshRoleAssignments: boolean = false): Observable<UserDetails> {
     return this.http.get<UserDetails>(`api/user/details?refreshRoleAssignments=${refreshRoleAssignments}`);
   }
+
+  // TODO use the ccd-case-ui-toolkit version of this instead
```
jamescatto-cgi, line 21: this function isn't exposed yet, but will be soon, so I have done this to pull user info access into a single place.

src/app/models/theming.model.ts
jamescatto-cgi: Thanks. As discussed, I'll leave this as is to avoid undue complexity

src/decentralisation/decentralised-casetype.ts
jamescatto-cgi: As discussed, I'll address this in a follow up PR

anthonydummer (dismissed):
LGTM

src/app/services/user/user.service.ts
jamescatto-cgi: As discussed, the `case ui toolkit code` can be used here when it has been exposed from that library in the near future. Additionally, there is a broader problem associated with accessing this `userInfo` from session storage, where it will be more stale than the `userInfo` stored in the redux store. However, it was agreed that this is a more general problem - ideally there should only be a single way to access this info - and that would be addressed at some future point. For this PR, this code should suffice.

connorpgpmcelroy (dismissed):
Approved based on PR comments, TODO needs to be actioned or produces ticket for the action

src/decentralisation/decentralised-casetype.ts
jamescatto-cgi: done in [this PR](https://github.com/hmcts/rpx-xui-webapp/pull/5361)

src/app/services/user/user.service.ts
jamescatto-cgi: This task is fairly involved and can be done independently: https://tools.hmcts.net/jira/browse/POFCC-380

anthonydummer (dismissed):

src/decentralisation/decentralised-redirect.service.ts
```diff
@@ -0,0 +1,78 @@
...
+      if (serviceMap && serviceId in serviceMap) {
+        const baseUrl = serviceMap[serviceId].baseUrl;
+
+        // prevent adding the baseURL if it has already been added
+        if (relativeUrl.startsWith(baseUrl)) {
+          return new URL(relativeUrl);
+        }
+
+        // try not to add a double slash
+        return baseUrl?.endsWith('/') || relativeUrl.startsWith('/')
```
Josh-HMCTS: I wonder if there is a different way of doing this, for example if a user currently sets
baseUrl = www.google.com/
relativeUrl = /help
That would effectively become www.google.com//help. 

Could we strip the / at the end of base and start of relative (if there is one) and then always add the / manually. 
Or we could say the services must follow the correct way of defining both base/relative whether that be 
/ on base and no / on relative
no / on base and / on relative
and leave the check as is just in case

Let me know your thoughts :)

jamescatto-cgi: Thanks, good idea, I've implemented the 'always strip' suggestion (see last commit)

Josh-HMCTS (dismissed):

connorpgpmcelroy (dismissed):

connorpgpmcelroy (approved):

Josh-HMCTS (approved):

PR #5320, opened 2026-07-22
CME-513: Fix pact for consumer: xui_webApp and providers:  acc_manageCaseAssignment, acc_manageCaseAssignment_Noc

anthonyb-cgi:
### Jira link

See [CME-513](https://tools.hmcts.net/jira/browse/CME-513)

### Change description

EXUI Consumer Pact tests were updated to support AAC Case Assignment provider verification by:

- Aligning expectations with provider behavior for `shared_with` `idam_id` fields to ensure contract compliance.
- Migrating interactions to the Pact V3 fluent DSL and matchers.
- Exercising the real consumer API client, including request/response transformation.
- Aligning provider names, state setup, case IDs, headers, paths, and query parameters with provider contracts.
- Relaxing matchers and removing response fields not used by the consumer.
- Adding `case_title`/`caseTitle` coverage for case-assignment responses.
- Updating NoC contracts to match the provider’s actual request and response payloads.

[CME-513]: https://hmcts.atlassian.net/browse/CME-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/test/pact/pact-tests/case-share/postCaseAssignments.spec.ts
```diff
@@ -1,64 +1,112 @@
...
-        },
-        willRespondWith: {
+          body: pactRequest,
+        })
+        .willRespondWith({
           status: 201,
-          headers: {
-            'Content-Type': 'application/json',
-          },
-          body: mockResponse,
```
kiran-yenigala-hmcts, line 41: why body is not validated here?

kiran-yenigala-hmcts (commented):
One comment added

api/test/pact/pact-tests/case-share/postCaseAssignments.spec.ts
anthonyb-cgi: Good question - as far as I can tell the real EXUI consumer API client doesn't use the provider response body field status_message when successfully calling it. Therefore, I have omitted validation of it from the pact (there used to be pact validation of the body previously when the consumer pact tests weren't testing against the real consumer API client).

However, if the response is rejected the real consumer API client does read the the API response so will check that.

anthonyb-cgi: Here is the new commit I made to cover consumer API consumption of the provider error response body:
https://github.com/hmcts/rpx-xui-webapp/pull/5320/changes/d71d6ace5ef259688bcd054663db7bdd0c906cf3

kiran-yenigala-hmcts (dismissed):
LGTM

RiteshHMCTS (dismissed):

RiteshHMCTS (dismissed):
LGTM

kiran-yenigala-hmcts (dismissed):
LGTM

connorpgpmcelroy (approved):
LGTM

RiteshHMCTS (approved):

andyg-cgi:
Thanks for the Pact cleanup here. One thing I think we should keep in the consumer contract is `case_type_id` on the NoC questions response.

The reason is that this field is not just display data. `getNoCQuestions` reads `data?.questions?.[0]?.case_type_id` and stores it in `req.session.nocCaseTypesByCaseId`. Later `validateNoCQuestions` and `submitNoC` use that cached case type to decide whether to route via the case-type-specific NoC service URL or fall back to the central case assignment URL.

With the current contract, the provider can stop returning `case_type_id` and the Pact would still pass, but the consumer would silently lose the decentralised routing signal. In practice that means the first call can look fine, while the later verify/submit calls go to the wrong NoC endpoint for case types such as Probate.

Suggested small fix:

```ts
const expectedResponse = {
  questions: eachLike({
    case_type_id: string('Probate'),
    order: integer(1),
    question_text: string('What is their Email?'),
    answer_field_type: {
      type: string('Email'),
      min: string('0'),
      max: string('10'),
      regular_expression: string('asdsa'),
    },
    question_id: string('QuestionId67745'),
  }),
};
```

And restore the assertion so the consumer test proves we consumed the field:

```ts
function assertResponse(returnedResponse: any) {
  expect(returnedResponse.questions.length).to.be.equal(1);
  expect(returnedResponse.questions[0].case_type_id).to.be.equal('Probate');
  expect(returnedResponse.questions[0].order).to.be.equal(1);
  // existing assertions...
}
```

If we want to make the test even more consumer-focused, we could also assert that the session cache is populated after the handler runs:

```ts
expect(req.session.nocCaseTypesByCaseId[caseId]).to.be.equal('Probate');
```

I am happy with the other removed response fields because they do not appear to drive consumer behaviour, but this one is part of the follow-on routing path so I think it should stay in the contract.

andyg-cgi (approved):

kiran-yenigala-hmcts (approved):
LGTM

balajisridharanhmcts (approved):

PR #5325, opened 2026-07-23
EXUI-4846 - Remove synthetic roles

Josh-HMCTS:
### Jira link

See [EXUI-4846](https://tools.hmcts.net/jira/browse/EXUI-4846)

### Change description

Remove the synthetic roles from user roles. 

### Testing done

Local testing didn't reveal any regression from removing this code


connorpgpmcelroy (approved):
As mentioned might be needed for some unknown reason so happy to approve but needs regression testing

RiteshHMCTS (approved):

PR #5418, opened 2026-09-09
pofcc-408 - add decentralisation guard to all routes where the event ID is passed in the URL

jamescatto-cgi:
### Jira link

https://tools.hmcts.net/jira/browse/POFCC-408

### Change description

Add router guard:

* to prevent access to routes when the event is decentralised
* to redirect the request to the decentralised event URL

- [ ] Yes
- [ ] No

### Checklist


src/cases/case-feature.routes.ts
```diff
@@ -75,6 +76,8 @@ export const ROUTES: Routes = [
           },
           {
             path: ':jid/:ctid/:eid',
+            canActivate: [DecentralisedEventGuard],
+            canActivateChild: [DecentralisedEventGuard],
```
jamescatto-cgi, line 80: I've added this because the `editorRouting `flows defined in `ccd-case-ui-toolkit` maybe should be redirected for editing. What do you think?

src/cases/case-feature.routes.ts
```diff
@@ -127,6 +130,7 @@ export const ROUTES: Routes = [
       },
       {
         path: 'case-details/:jurisdiction/:caseType/:cid',
+        canActivateChild: [DecentralisedEventGuard],
```
jamescatto-cgi, line 133: although this doesn't include the eventID parameter, some of the child routes do (defined in `caseViewRouting` in `ccd-case-ui-toolkit`), so this will redirect any that do. Is this correct?

src/decentralisation/decentralised-event.guard.ts
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        if (event) {
+          const baseUrl = this.decentralisedRedirectService.getBaseUrl(caseType);
+
+          if (baseUrl) {
+            this.decentralisedRedirectService.redirectEvent(baseUrl, event);
+            return false;
+          } else {
+            // fail fast since decentralised events should have the required configuration
```
jamescatto-cgi, line 45: Throwing an error here seems like a good idea, because we shouldn't allow normal navigation when a decentralised event is occurring and all the other required parameters are provided. Ideally, we would throw an error for any decentralised event that doesn't have the required parameters. But I haven't done that because the guard is placed on top level routes where only some of the children may need to be redirected (see routing config). That saves us having to reorganise the code, but is arguably less correct and so more buggy. WDYT?

jamescatto-cgi (commented):
@danlysiak as discussed, what do you think of this approach?

src/cases/containers/case-home/case-home.component.ts
```diff
@@ -18,6 +18,7 @@ import { HeaderComponent } from '../../../app/components';
 import * as fromRoot from '../../../app/store';
 import * as fromFeature from '../../store';
 import { DecentralisedRedirectService } from '../../../decentralisation/decentralised-redirect.service';
+import { DecentralisedEvent } from 'src/decentralisation/decentralised-event';
```
connorpgpmcelroy: src should be ../../../ as above

src/cases/store/effects/case-create.effects.ts
```diff
@@ -7,6 +7,7 @@ import { LoggerService } from '../../../app/services/logger/logger.service';
 import * as fromRoot from '../../../app/store';
 import * as fromActions from '../actions';
 import { DecentralisedRedirectService } from '../../../decentralisation/decentralised-redirect.service';
+import { DecentralisedEvent } from 'src/decentralisation/decentralised-event';
```
connorpgpmcelroy: As above

connorpgpmcelroy (changes requested):
Only two minor changes requested

connorpgpmcelroy (approved):

RiteshHMCTS (approved):

PR #5421, opened 2026-09-10
EXUI-253: Don't redirect CCD requests to the CCD gateway

chrisjones-hmcts:
### Jira link

See [EXUI-253](https://tools.hmcts.net/jira/browse/EXUI-253)

### Change description

Routes CCD Data Store requests directly to the Data Store API instead of through the gateway, preserving the required path rewrites and search query parameters. Rewrites aggregated caseworker URLs using the authenticated user id.

- [ ] Yes
- [x] No

[EXUI-253]: https://hmcts-sandbox4.atlassian.net/browse/EXUI-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/proxy.config.ts
```diff
@@ -68,18 +68,24 @@ export const initProxy = (app: Express) => {
...
-    filter: ['!/data/internal/searchCases'],
     rewrite: false,
-    source: ['/print', '/data'],
+    source: '/print',
     target: getConfigValue(SERVICES_CCD_COMPONENT_API_PATH),
   });
 
+  applyProxy(app, {
+    filter: ['!/data/internal/searchCases'],
+    rewrite: true,
```
jamescatto-cgi, line 84: Just for my understanding, does this need a rewriteUrl?

api/proxy.config.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    sandbox = sinon.createSandbox();
+    applyProxyStub = sandbox.stub(middlewareProxy, 'applyProxy');
+    getConfigValueStub = sandbox.stub(configuration, 'getConfigValue').callsFake((reference: string) => reference);
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('routes Data Store-backed CCD paths directly to the Data Store API', () => {
```
jamescatto-cgi, line 24: this test covers a lot of ground; it may be nicer to split them into more manageable chunks which would also help describe each test

jamescatto-cgi (dismissed):

api/proxy.config.ts
chrisjones-hmcts: a rewriteUrl is not needed here because our middlware removes the /data prefix for that proxy before forwarding the request to ccd data store in the expected /internal/ path.

api/proxy.config.spec.ts
chrisjones-hmcts: good suggestion thanks - ive split this up and increased the unit test coverage

jamescatto-cgi (approved):
LGTM

balajisridharanhmcts (approved):

PR #5423, opened 2026-09-10
Task/linked hearings display changes

connorpgpmcelroy:
### Jira link

https://hmcts.atlassian.net/browse/EXUI-4631
https://hmcts.atlassian.net/browse/EXUI-4632
https://hmcts.atlassian.net/browse/EXUI-4633
https://hmcts.atlassian.net/browse/EXUI-4634

### Change description

Added hearing ID to linked hearing displays
Ensured current case and hearing is pre-selected
Case hearings can only be selected if case link reasons includes linked for a hearing

### Testing done

Confirmed display looks correct and only hearings with correct case link reason can be used

- [ ] Yes
- [ x ] No


RiteshHMCTS (dismissed):

src/hearings/containers/linked-hearings/linked-hearings-with-case/linked-hearings-with-case.component.ts
```diff
@@ -137,38 +158,61 @@ export class LinkedHearingsWithCaseComponent implements OnInit, OnDestroy {
...
-          isSelected: this.shouldSelected(hearingInfo),
+          isSelected: this.shouldSelected(hearingInfo, caseRef),
         })
       )
     );
   }
 
-  public shouldSelected(hearingInfo): boolean {
+  public shouldSelected(hearingInfo: HearingDetailModel, caseRef?: string): boolean {
+    if (caseRef && this.isCurrentHearing(caseRef, hearingInfo.hearingID)) {
```
andyg-cgi, line 168: Could we also require another hearing to be selected in the fresh-link submission path? With the current hearing now selected automatically, the existing validator considers the untouched form valid, so Continue advances without another hearing being chosen.

`isGetHearingsSelected()` already excludes the current hearing. In the non-manage branch of `onSubmit()`, we can reuse it without changing the existing error handling:

```diff
- if (this.linkHearingForm.valid) {
+ if (this.linkHearingForm.valid && this.isGetHearingsSelected()) {
    this.saveLinkedHearingInfo();
```

I checked this guard using the component's extracted methods and real Angular forms: current-only is rejected, while current plus another hearing is accepted with both IDs preserved in the form state. This check did not exercise the dispatched request or persistence. A regression test for those two states would help protect the change.

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+function buildHearingLinkSessionCookies(
+  baseUrl = process.env.TEST_URL?.trim() || 'https://manage-case.aat.platform.hmcts.net'
+): Cookie[] {
+  const origin = new URL(baseUrl).origin;
+  const secure = origin.startsWith('https:');
+  return [
+    {
+      name: '__userid__',
+      value: 'hearing-link-local-user',
+      url: origin,
```
andyg-cgi, line 123: The cookie values can stay as they are; the type needs to describe the input to `addCookies()`. `Cookie` describes cookies returned by the browser, so TypeScript currently rejects `url` on all three objects.

We can derive the input type directly from Playwright. Add this type import and alias in this spec and `hearingJourneySetup.helper.ts`, replacing the now-unused `Cookie` import:

```ts
import type { BrowserContext } from '@playwright/test';

type SessionCookies = Parameters<BrowserContext['addCookies']>[0];
```

Then change the builder return annotation:

```diff
- ): Cookie[] {
+ ): SessionCookies {
```

And change both helper option declarations:

```diff
- sessionCookies?: Cookie[];
+ sessionCookies?: SessionCookies;
```

This keeps the types aligned with the API we actually call, without changing cookie behavior.

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
```diff
@@ -162,43 +162,61 @@ export class HearingsJourneyPage {
...
+        }
+
+        // Guards against a mis-targeted selection silently adding a different court: the tag the
+        // page added must be the venue that was searched for.
+        if (!newSelectedVenue.includes(venueSearchTerm)) {
+          throw new Error(`Expected the added venue to match "${venueSearchTerm}", but the page added "${newSelectedVenue}".`);
+        }
+
+        return newSelectedVenue;
+      } catch (error) {
```
andyg-cgi: Could we keep the original failure visible here? The outer loop catches every error, including no results and an incorrect venue being added. A later successful attempt can hide the first failure; a controlled check reproduced that behavior.

A small option is to keep the existing selection steps and diagnostics, but run them once. Rename the helper to `selectVenue` and update its caller:

```ts
return this.selectVenue(venueSearchTerm, seededVenueCount);
```

The helper would be:

```ts
private async selectVenue(venueSearchTerm: string, seededVenueCount: number): Promise<string> {
  await this.hearingVenue.fill('');
  await this.hearingVenue.pressSequentially(venueSearchTerm);
  const venueOption = this.venueOptionsMatching(venueSearchTerm).first();
  try {
    await venueOption.waitFor({ state: 'visible', timeout: 30_000 });
  } catch (error) {
    if (await this.venueNoResultsOption.isVisible()) {
      throw new Error(`Location search for "${venueSearchTerm}" returned "No results found".`);
    }
    throw error;
  }
  await this.hearingVenue.press('Enter');
  await this.addLocationsButton.click();
  const addedVenueTag = this.selectedVenueTags.nth(seededVenueCount);
  await addedVenueTag.waitFor({ state: 'visible', timeout: 30_000 });
  const newSelectedVenue = await this.venueTagName(addedVenueTag);
  if (!newSelectedVenue) {
    throw new Error(`Venue selected for "${venueSearchTerm}" did not expose a location name.`);
  }
  if (!newSelectedVenue.includes(venueSearchTerm)) {
    throw new Error(
      `Expected the added venue to match "${venueSearchTerm}", but the page added "${newSelectedVenue}".`,
    );
  }
  return newSelectedVenue;
}
```

A fake-driven check of this version preserves the successful result and propagates both an action failure and a wrong-venue error without replay. This is not a browser validation of the Enter interaction. If a specific transient state needs retrying, we can target that state once trace evidence identifies it.

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+  selectOrderedLinkedHearings,
+} from '../../helpers';
+import type { HearingsTabPage } from '../../../E2E/page-objects/pages/exui/hearingsTab.po';
+import type { CaseDetailsPage } from '../../../E2E/page-objects/pages/exui/caseDetails.po';
+
+const linkedCasesWithHearingReason = [
+  {
+    caseReference: '1611573453599537',
+    caseName: 'redacted redacted - appellantnamefordisplay',
+    reasonsForLink: [CASE_LINKING_SECONDARY_REASON_CODE, CASE_LINKING_REASON_CODE],
```
andyg-cgi: Could we include one case without the hearing-link reason? The current fixture contains only an eligible case, so the row/content assertions also pass with the previous filtering implementation. “Not child” is absent from the supplied data, which means that assertion does not exercise exclusion.

Add this case to the mock response alongside the eligible case:

```ts
{
  caseReference: '1652112127295262',
  caseName: 'Case linked for consolidation only',
  reasonsForLink: [CASE_LINKING_REASON_CODE], // CLRC015, without CLRC017
}
```

Then assert its row is absent using the existing `linkedCases` locator:

```ts
await expect(
  linkedCases.filter({ hasText: 'Case linked for consolidation only' })
).toHaveCount(0);
```

I checked the data against the source mapping: the PR excludes this case and the old implementation retains it. A browser run with the filter temporarily removed should then fail this assertion.

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+    { sessionCookies: hearingLinkSessionCookies }
+  );
+
+  await hearingsTabPage.waitForReady(LISTED_HEARING_SCENARIO.hearingId, 'link');
+  await expect(hearingsTabPage.requestHearingButton).toBeVisible({ timeout: 60_000 });
+  await hearingsTabPage.openLinkHearing(LISTED_HEARING_SCENARIO.hearingId);
+  await expect(page.getByRole('heading', { name: /which hearings should be linked\?/i })).toBeVisible();
+}
+
+async function setHearingPositions(page: Page, positions: string[]): Promise<void> {
```
andyg-cgi: Could we reuse the existing hearing-test structure for the setup in this file? Moving the hearing payloads into the mock module and the reusable journey/session setup into the existing helpers would make the scenarios easier to follow and keep future setup changes in one place.

For example, this repeated ordering interaction can live on `HearingsTabPage`, which already exposes `hearingOrderSelects`:

```ts
async setHearingPositions(positions: string[]): Promise<void> {
  for (const [index, position] of positions.entries()) {
    await this.hearingOrderSelects.nth(index).selectOption(position);
  }
}
```

The spec can then use:

```ts
await hearingsTabPage.setHearingPositions(['1', '2', '3']);
```

The assertions should stay visible in the spec. The duplicate-position rejection scenario would also fit the existing `.negative.spec.ts` convention. This can use the current modules rather than introducing another abstraction layer.

src/hearings/containers/linked-hearings/linked-hearings-with-case/linked-hearings-with-case.component.ts
```diff
@@ -137,38 +158,61 @@ export class LinkedHearingsWithCaseComponent implements OnInit, OnDestroy {
...
+  }
+
   public initForm(): void {
     this.linkHearingForm = this.fb.group(
       {
         linkedCasesWithHearings: this.getCasesFormArray,
       },
       { validator: this.validators.validateLinkedHearings() }
     );
+    this.setCurrentHearingSelected();
```
andyg-cgi: Optional simplification: could we remove this second selection pass? Form creation already passes the case reference into `shouldSelected()`, which selects the current hearing.

```diff
- this.setCurrentHearingSelected();
```

The `setCurrentHearingSelected()` method and its comments can then be removed too. I checked the form values with this call disabled using real Angular forms; they were identical. That would leave one place responsible for the initial selection.

andyg-cgi (commented):
Review comments with suggested fixes, checked against head `1e6d745c6`.

Validation of the suggestions: selection guard checked with real Angular forms; cookie input typing checked against installed Playwright types; filtering data compared against the base and PR mapping; venue and ordering helper behavior checked with controlled fakes. These checks do not replace full integration/browser validation. The earlier focused review also passed 32 Angular unit tests and 8 mock-builder tests.

The selection, typing and test-evidence comments are the main follow-ups. The final simplification is optional. No source changes have been pushed. Review prepared with AI assistance and independently checked.

src/hearings/containers/linked-hearings/linked-hearings-with-case/linked-hearings-with-case.component.ts
connorpgpmcelroy: I will check this as I can see what you mean. this.setCurrentHearingSelected was made because manage links didn't have the current hearing as selected for some reason but you're right, it appears as though it should be. Might be a problem specific to managing links as opposed to linking hearings but I'll check this

src/hearings/containers/linked-hearings/linked-hearings-with-case/linked-hearings-with-case.component.ts
connorpgpmcelroy: I will check the continue on linking hearings because yes, the current hearing being selected should not allow the user to continue. I will check the suggested change

playwright_tests_new/E2E/page-objects/pages/exui/hearingsJourney.po.ts
connorpgpmcelroy: This isn't ideal though, we are then relying on success immediately. Ideally we want to try each test multiple times or builds can become reliant and flaky. Also this test wasn't made by me (although may have been edited slightly to allow slight discrepancies after hearing changes made) so could you create a follow on ticket to check/evaluate this?

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
connorpgpmcelroy: Yes, that's fine, will do

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
connorpgpmcelroy: I'll make that change then, hopefully works

playwright_tests_new/integration/test/hearings/hearingLink.positive.spec.ts
connorpgpmcelroy: I didn't create these playwright tests so I don't know why we didn't reuse the structure but again, could this be done in a testing ticket follow on or is it the case that without this change the functionality won't be correctly being tested?

connorpgpmcelroy (commented):
Added responses. Most comments I will check and modify code for but some about playwright test setup I am not sure I/Codex should edit as it was made for the automation first outlook

andyg-cgi (dismissed):

RiteshHMCTS (approved):

andyg-cgi (approved):

